PR #12, opened 2022-10-04
Update ledger footprint in simulateTransaction to be base64-xdr

paulbellamy:


Shaptic (approved):
Isn't this a breaking change? :stuck_out_tongue_closed_eyes: you should probably keep a `-beta.[N]` suffix to stay semver-y.

leighmcculloch:
> you should probably keep a `-beta.[N]` suffix to stay semver-y.

We should probably be only releasing v0.0x builds, that's what we're doing in the Rust crates.

Shaptic:
or that yeah

PR #30, opened 2022-12-20
Add support for the `getEvents()` RPC method.

Shaptic:


src/server.ts
```diff
@@ -185,7 +189,59 @@ export class Server {
   }
 
   /**
-   * Submit a trial contract invocation to get back return values, expected ledger footprint, and expected costs.
+   * Fetches all events that match a given set of filters.
+   *
+   * The given filters are combined only in a logical AND fashion, and all of
```
paulbellamy, line 194:
```suggestion
   * The given filters are combined only in a logical OR fashion, and all of
```

src/soroban_rpc.ts
```diff
@@ -64,6 +64,28 @@ export namespace SorobanRpc {
...
   }
 
+  export interface EventFilter {
+    type?: string;
+    contractIds?: string[];
+    topics?: string[][];
+  }
+
+  export interface GetEventsResponse {
+    events?: EventResponse[];
```
paulbellamy, line 74: pagination?

paulbellamy: Let's either implement it, or let's un-document it if that is easier for now. (There was a comment that says it is supported)

src/server.ts
Shaptic: Nice, didn't realize.

src/soroban_rpc.ts
Shaptic: Added in latest push :+1:

paulbellamy (approved):

PR #31, opened 2022-12-20
Add pre-commit hook to lint test files.

Shaptic:
Two things:
 * run `prettier --write test/**/*.js`
 * add a pre-commit hook to `lint-staged` to do the same thing


paulbellamy:
Fine for now, but we should stop using husky. It doesn't work right with git. We should check the formatting with a CI action.

paulbellamy (approved):

PR #36, opened 2023-01-18
changed startLedger and endLedger of getEvents to be string types 

sreuland:
currently getting error response from rpc server for `server.getEvents` invocations:
```
{
  code: -32602,
  message: 'invalid parameters: [-32602] invalid parameters'
}
```

the rpc server jsonrpc handler expects `startLedger and `endLedger`  in `getEvents:params` to be integers expressed as string types, also saw the same error when sending `params:[{...}]` , getEvents wasn't able to unmarshal the params array, it only worked as `params:{}`

Closes https://github.com/stellar/js-soroban-client/issues/38

### DIFF
--- a/src/jsonrpc.ts
+++ b/src/jsonrpc.ts
@@ -33,7 +33,12 @@ export async function post<T>(
   ...params: any
 ): Promise<T> {
   let jsonParams = params;
-  if (params.length === 1) {
+  if (
+    params.length === 1 &&
+    typeof params[0] === "object" && params[0] !== null
+  ) {
+    // if only one param and it's a js object, then pass the object.
+    // otherwise rpc server throws unmarshal error on jsonrpc params.
     jsonParams = params[0];
   }
   const response = await axios.post<Response<T>>(url, {
--- a/src/jsonrpc.ts
+++ b/src/jsonrpc.ts
@@ -27,26 +27,42 @@ export interface Error<E = any> {
   data?: E;
 }
 
+/**
+ * Sends the jsonrpc 'params' as an array.
+ */
 export async function post<T>(
   url: string,
   method: string,
   ...params: any
 ): Promise<T> {
-  let jsonParams = params;
-  if (
-    params.length === 1 &&
-    typeof params[0] === "object" && params[0] !== null
-  ) {
-    // if only one param and it's a js object, then pass the object.
-    // otherwise rpc server throws unmarshal error on jsonrpc params.
-    jsonParams = params[0];
-  }
   const response = await axios.post<Response<T>>(url, {
     jsonrpc: "2.0",
     // TODO: Generate a unique request id
     id: 1,
     method,
-    params: jsonParams,
+    params,
+  });
+  if (hasOwnProperty(response.data, "error")) {
+    throw response.data.error;
+  } else {
+    return response.data?.result;
+  }
+}
+
+/**
+ * Sends the jsonrpc 'params' as the single 'param' obj, no array wrapper is applied.
+ */
+export async function postObject<T>(
+  url: string,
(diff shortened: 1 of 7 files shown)


src/jsonrpc.ts
```diff
@@ -32,12 +32,16 @@ export async function post<T>(
   method: string,
   ...params: any
 ): Promise<T> {
+  let jsonParams = params;
+  if (params.length === 1) {
+    jsonParams = params[0];
+  }
```
paulbellamy, line 38: Oh, is this trying to handle when `params` is an object? Or, I'm not sure what this is doing...

sreuland: @paulbellamy , yes, I cleaned up the code here, i ran into two errors when trying the client against rpc server:
* params.endLedger and params.startLedger need to be strings
* params needs to be object not array for getEvents, otherwise jsronrpc had unmarshal error on it.

paulbellamy: Ok, but as it is, I suspect this change might break methods with a single parameter. For example, I pass `[xdr.ScVal(...)]`, but then that get's converted to just `xdr.ScVal(...)`, and the call fails. I expect object/array thing is because the pagination param at the end is required (not optional, bug in the docs). Was that what you were hitting?

Could we just do the startLedger/endLedger change?

sreuland: will still get error due to the params being an array, was getting error from rpc `getEvents`server handler, 

```
 "error": {
        "code": -32602,
        "message": "invalid parameters: [-32602] invalid parameters"
    }
```

the jrpc2 lib has json unmarsal errors with this `params` json when it introspects and tries to marshal it into [GetEventsRequest](https://github.com/stellar/soroban-tools/blob/main/cmd/soroban-rpc/internal/methods/get_events.go#L43) struct, both for number types on startLedger/endLedger, and the params being an array with a single object. 

the jsonrpc spec allows array or object type for the `params`, and i saw jrpc2 tries to handle both, but ends up in error anyways which I didn't step through deeper to see why. but it seems that `params:[jsObj]` and `params:jsObj` are both valid forms and should resolve to same `jsObj` , which covers your example such as `[xdr.ScVal(...)]` 

Looking at the other server.ts methods, none are passing arrays with single js objects yet, they are mostly passing arrays with single scalar value, and the [change introduced here ](https://github.com/stellar/js-soroban-client/pull/36/files#diff-f4489a68857458c653600f8a11e8a2bac870bd80723a9bdc89676fd5bdae49d8R37), would leave that as-is. 

I would need to execute all the server.ts methods from this pr against current rpc server manually to validate your concern.  We need an integration test between js client and rpc to verify these two are compliant with each other, maybe can automate by adding those calls to the existing integration `client_headers_test.js`? In parallel, I think system-test will start to establish that coverage with https://github.com/stellar/system-test/issues/5

How about i just create a `jsonrpc.postObject()` method and have getEvents use that form, to isolate this for now to getEvents, which can confirm it will work in that scope.

sreuland: I created issue #38 , has curl examples of the getEvents call with `params` formats that error and work.

paulbellamy: Oh that's true it would only affect soroban-rpc methods with a single arg. I was mixing that up with contract fns with a single arg. Separate `postObject` used by `getEvents` to be more explicit would be good, IMO.

sreuland: ok, added `jsonrpc.postObject` - https://github.com/stellar/js-soroban-client/pull/36/commits/4cb61f2e57386cfe62a60e690de041130e7e0503

paulbellamy (approved):

Issue #39, opened 2023-01-21
Example in docs doesn't quite work


deian:
I'm trying to get the counter example working locally but hit a few bugs in the docs (and maybe elsewhere).

Setup:
- I'm running a local network following: https://soroban.stellar.org/docs/tutorials/deploy-to-local-network
- The contract I deployed is the counter example from: https://soroban.stellar.org/docs/examples/storing-data
- Interacting with the contract using the rust cli works just fine
- Using the js client fails
   - I followed both the readme and examples in this repo; both have code that is a bit stale afaict
   - Here is the thing I ended up with after fixing some of the bugs in the examples (the `XXX` comments are the changes I made to make the script get it to at least send transactions):
 
```javascript
const SorobanClient = require('soroban-client');

// XXX need http to not crash with unsecure connection error
const server = new SorobanClient.Server('http://localhost:8000/soroban/rpc', { allowHttp: true });
const secretKey = 'SBKROBEITSLZJDTK6W5A7OMDAO6X47Q2E4B4FS5ST5NCJ2USDCBJNOTK';
const keypair = SorobanClient.Keypair.fromSecret(secretKey);
const publicKey = keypair.publicKey();
const contractId = '74aeaf0a3d4a4c3419af98b51cce1fb3e8d01b7fdaa60f8b17c080d0a613f232';

(async function main() {
  // // XXX getAccount (as used in readme.md) doesn't return an Account so build will fail
  const _account = await server.getAccount(publicKey);
  const account = new SorobanClient.Account(_account.id, _account.sequence);
  /// XXX: loadAccount (as used https://github.com/stellar/js-soroban-client/blob/main/docs/reference/examples.md) is not a method on server 

  // Right now, this is just the default fee for this example.
  const fee = 100;

  const contract = new SorobanClient.Contract(contractId);

  const transaction = new SorobanClient.TransactionBuilder(account, {
    fee,
    // XXX I think this should be STANDALONE not TESTNET to work with 
    // local network as described <https://soroban.stellar.org/docs/tutorials/deploy-to-local-network>
    networkPassphrase: SorobanClient.Networks.STANDALONE
  })
  // Add a contract.increment soroban contract invocation operation
    .addOperation(contract.call("increment"))
  // Make this transaction valid for the next 30 seconds only
    .setTimeout(30)
  // Uncomment to add a memo (https://developers.stellar.org/docs/glossary/transactions/)
  // .addMemo(SorobanClient.Memo.text('Hello world!'))
[...]

PR #45, opened 2023-01-30
Add `server.prepareTransaction`

paulbellamy:
Adds a `server.prepareTransaction` method, which calls `simulateTransaction` then adds the resulting footprint into the transaction.

Depends on https://github.com/stellar/soroban-tools/pull/366

Fixes https://github.com/stellar/js-soroban-client/issues/39

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -391,9 +391,12 @@ export class Server {
    */
   public async prepareTransaction(
     transaction: Transaction | FeeBumpTransaction,
+    networkPassphrase?: string,
   ): Promise<Transaction | FeeBumpTransaction> {
     let [{ passphrase }, { footprint }] = await Promise.all([
-      this.getNetwork(),
+      networkPassphrase
+        ? Promise.resolve({ passphrase: networkPassphrase })
+        : this.getNetwork(),
       this.simulateTransaction(transaction),
     ])
     return addFootprint(transaction, passphrase, footprint);
--- a/src/server.ts
+++ b/src/server.ts
@@ -274,7 +274,11 @@ export class Server {
    * Fetches metadata about the network which Soroban-RPC is connected to.
    *
    * @example
-   * server.getNetwork();
+   * server.getNetwork().then(network => {
+   *   console.log("friendbotUrl:", network.friendbotUrl);
+   *   console.log("passphrase:", network.passphrase);
+   *   console.log("protocolVersion:", network.protocolVersion);
+   * });
    *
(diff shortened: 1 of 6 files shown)


src/soroban_rpc.ts
```diff
@@ -39,6 +39,17 @@ export namespace SorobanRpc {
...
+  /* Response for jsonrpc method `getNetwork`
+   */
+  export interface GetNetworkResponse {
+    friendbotUrl?: string;
+    latestLedgerHash: string;
+    latestLedgerSeq: number;
+    latestLedgerTime: string;
+    passphrase: string;
+    protocolVersion: string;
+  }
```
Shaptic, line 51: Can we reconcile this with the return values in [the doc](https://docs.google.com/document/d/1TZUDgo_3zPz7TiPMMHVW_mtogjLyPL0plvzGMsxSz6A/edit#heading=h.3zrwisrq3rg3)? It's hard to know where the source of truth is when reviewing fields/types/etc.

src/server.ts
```diff
@@ -329,6 +342,66 @@ export class Server {
...
+   *
+   * @param {Transaction | FeeBumpTransaction} transaction - The transaction to
+   *    prepare. It should include exactly one operation, which must be a
+   *    {@link InvokeHostFunctionOp}. Any provided footprint will be overwritten.
+   * @returns {Promise<Transaction | FeeBumpTransaction>} Returns a copy of the
+   *    transaction, with the expected ledger footprint added.
+   */
+  public async prepareTransaction(
+    transaction: Transaction | FeeBumpTransaction,
+    networkPassphrase?: string,
```
Shaptic, line 394: not documented? we should clarify that this will be an `async` call if this isn't passed

src/server.ts
```diff
@@ -271,6 +270,20 @@ export class Server {
     });
   }
 
+  /**
+   * Fetches metadata about the network which Soroban-RPC is connected to.
+   *
+   * @example
+   * server.getNetwork();
```
Shaptic, line 277: lol

Shaptic (approved):
LGTM, just some small doc notes :+1:

src/soroban_rpc.ts
paulbellamy: oh duh. Yeah. This response is wrong, sorry!

src/server.ts
paulbellamy: I mean... I'll add some console.logs of the outputs for this example

PR #46, opened 2023-02-01
Implement `server.requestAirdrop` method to create & fund a new testnet account

paulbellamy:


test/unit/server/request_airdrop_test.js
```diff
@@ -0,0 +1,153 @@
...
+      .then(function(_) {
+        done(new Error("Should have thrown"));
+      })
+      .catch(function(err) {
+        expect(err.message).to.be.equal("No friendbot URL configured for current network");
+        done();
+      });
+  });
+
+  it('throws if there the request fails', function(done) {
```
Shaptic, line 133: typo?
```suggestion
  it('throws if the request fails', function(done) {
```

test/unit/server/request_airdrop_test.js
```diff
@@ -0,0 +1,153 @@
...
+      .requestAirdrop(accountId)
+      .then(function(_) {
+        done(new Error("Should have thrown"));
+      })
+      .catch(function(err) {
+        expect(err.message).to.be.equal("Request failed");
+        done();
+      });
+  });
+});
```
Shaptic, line 153: sweet test coverage, nice job :clap:

Shaptic (commented):
solid, just a few minor nits and a design question

src/server.ts
```diff
@@ -467,6 +468,45 @@ export class Server {
...
+      const network = await this.getNetwork();
+      friendbotUrl = network.friendbotUrl;
+    }
+    if (!friendbotUrl) {
+      throw new Error("No friendbot URL configured for current network");
+    }
+    const response = await AxiosClient.post<Friendbot.Response>(
+      `${friendbotUrl}?addr=${encodeURIComponent(account)}`
+    );
+    return !!response.data?.successful;
```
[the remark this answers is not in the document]
paulbellamy: booleanize

test/unit/server/get_network_test.js
```diff
@@ -0,0 +1,43 @@
...
+      passphrase: 'Soroban Testnet ; December 2018',
+      protocolVersion: 20,
+    };
+    this.axiosMock
+      .expects('post')
+      .withArgs(
+        serverUrl,
+        {
+          jsonrpc: '2.0',
+          id: 1,
```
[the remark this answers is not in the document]
paulbellamy: lmaoyolo

src/server.ts
```diff
@@ -467,6 +468,45 @@ export class Server {
       transaction.toXDR(),
     );
   }
+
+  /**
+   * Use the friendbot faucet to create and fund a new account. If friendbot is
+   * not configured on this network, this method will throw an error. The
+   * method will return true if the account was created and funded successfully,
+   * or false if the account already existed. If the request fails, this method
```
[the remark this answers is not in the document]
paulbellamy: Yeah, I wanted to do that. Unfortunately, friendbot only returns the account if it is newly created, not if it already exists.

Though, actually, I realized this is wrong as well. If the account already exists friendbot throws a `400 Bad Request`

src/server.ts
```diff
@@ -487,22 +494,29 @@ export class Server {
...
+        `${friendbotUrl}?addr=${encodeURIComponent(account)}`,
+      );
+      return new Account(account, "1");
+    } catch (error) {
+      if (error.response?.status === 400) {
+        // Account already exists, load the sequence number
+        return this.getAccount(account);
+      }
+      throw error;
+    }
```
Shaptic, line 518: Imma be annoying for a sec for my own edification. I know you mentioned preference for try/catch in the dApp, but are we / should we do a similar thing here? what about chaining?

```ts
return AxiosClient
    .post<Friendbot.Response>(`...`)
    .then((account) => new Account(account, account.sequence)
    .catch((err) => {
        if (err.response?.status === 400) {
            return this.getAccount(account);
        }
        return err;
    });
```

Shaptic: Also, note two additional differences in the above:

 - there's no need to `await` the friendbot request since we return a Promise, anyway (right?)
 - the sequence number isn't necessarily 1 after funding so we need to check the `post` retval (I think?)

test/unit/server/request_airdrop_test.js
```diff
@@ -33,7 +35,7 @@ describe('Server#requestAirdrop', function() {
 
   it('returns true when the account is created', function(done) {
     const friendbotUrl = 'https://friendbot.stellar.org';
-    const accountId = "GBZC6Y2Y7Q3ZQ2Y4QZJ2XZ3Z5YXZ6Z7Z2Y4QZJ2XZ3Z5YXZ6Z7Z2Y4";
+    const accountId = 'GBZXN7PIRZGNMHGA7MUUUF4GWPY5AYPV6LY4UV2GL6VJGIQRXFDNMADI';
```
Shaptic, line 38: Can you review/merge #31 before/after this so we get consistency? Otherwise, this will be dependent on who checks out the repo, right?

src/server.ts
Shaptic: (Can you add a test for the second bullet point, i.e. ensure the friendbot-funded return value matches the getAccount return value?)

src/server.ts
```diff
@@ -467,6 +475,48 @@ export class Server {
       transaction.toXDR(),
     );
   }
+
+  /**
+   * Use the friendbot faucet to create and fund a new account. If friendbot is
+   * not configured on this network, this method will throw an error. The
+   * method will return an Account object for the created account. If the
+   * request fails, this method will throw an error.
```
Shaptic, line 483: also reminder to update the docstring for the new "retrieve on error" behavior

src/server.ts
paulbellamy: Why wouldn't the sequence number on a newly created account be 1?

paulbellamy: and re: chaining. We haven't needed to do either here, yet. Generally I've been using async/await. But whatever you prefer.

src/server.ts
paulbellamy: How is that different than what this says?

src/server.ts
tomerweller: > Why wouldn't the sequence number on a newly created account be 1?

New account sequence number is determined by the sequence number of the ledger in which it's created (IIRC 32 bits left shifted)

Shaptic: I didn't know the details, I just knew it wasn't 1. Thanks Tomer! (tl;dr: this avoids `MergeAccountOp` wonkery)

With that in mind, I do think we have to use the return value from `getAccount` to build the return value, and since it's async, `.then()`ifying it also makes sense to make the method return asap (whether or not that's done via `try`/`catch` or `.error()` is up to you; I was just asking to learn).

src/server.ts
Shaptic: I misread it somehow; you're right :+1: maybe we can say `"...for the created account, returning it if it already existed."

src/server.ts
paulbellamy: Cool, TIL.

Shaptic (approved):
I can't believe friendbot doesn't return the new account.. sorry you had to figure out `findCreatedAccountSequenceInTransactionMeta` :sweat_smile:

paulbellamy:
Yeah, returning the account object seemed far too intuitive 😭

PR #47, opened 2023-02-02
Updates for auth-next

paulbellamy:
Depends on:
- [x] https://github.com/stellar/js-stellar-base/pull/570
- [x] https://github.com/stellar/soroban-tools/issues/348
- [x] Add helpers to build contract and account addresses: https://github.com/stellar/js-stellar-base/pull/572

Part of https://github.com/stellar/soroban-tools/issues/350


tsachiherman (commented):
In the code, you've added quite many `todo`. Is that good enough to be released as is?

src/server.ts
```diff
@@ -408,13 +407,19 @@ export class Server {
...
       this.simulateTransaction(transaction),
     ]);
-    return addFootprint(transaction, passphrase, footprint);
+    if (error) {
+      throw error;
+    }
+    if (!results) {
+      throw new Error("transaction simulation failed");
+    }
+    return prepareTransaction(transaction, passphrase, results);
```
sreuland, line 422: at first I read this as recursive, just for minor consideration, alias or rename the imported `prepareTransaction` to something like `assembleTransaction`

src/transaction.ts
```diff
@@ -0,0 +1,78 @@
...
+  Account,
+  FeeBumpTransaction,
+  Operation,
+  Transaction,
+  TransactionBuilder,
+  xdr,
+} from "stellar-base";
+
+// TODO: Transaction is immutable, so we need to re-build it here. :(
+export function prepareTransaction(
```
sreuland, line 11: a unit test demonstrating input/output would be nice if feasible, the visibility from that could also double as a developer reference for latest doc on `SimulateTransactionResponse.results`.

sreuland (approved):
looks good, approved, one suggestion on unit test coverage if possible, thanks!

paulbellamy:
The `TODO` comments in `transaction.ts` are actually not new. Previously they were copy-pasted between the example dapp and soroban-react packages. This change moves them here to standardize the code, and then we can work on improving it. It definitely needs improvement, but it's also better than what we have on offer today (nothing).

tsachiherman (approved):

PR #49, opened 2023-02-11
update getEvents request

tamirms:
This commit updates the js soroban client to match the changes to the getEvents endpoint which are documented here: https://github.com/stellar/soroban-docs/pull/293

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -21,7 +21,7 @@ export const SUBMIT_TRANSACTION_TIMEOUT = 60 * 1000;
 
 export interface GetEventsRequest {
   startLedger?: number;
-  filters?: SorobanRpc.EventFilter[];
+  filters: SorobanRpc.EventFilter[];
   cursor?: string;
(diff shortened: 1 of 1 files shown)


src/server.ts
```diff
@@ -19,6 +19,13 @@ import { SorobanRpc } from "./soroban_rpc";
 
 export const SUBMIT_TRANSACTION_TIMEOUT = 60 * 1000;
 
+export interface GetEventsRequest {
+  startLedger?: number;
+  filters?: SorobanRpc.EventFilter[];
```
paulbellamy, line 24: startLedger and filters are still both required here, afaik (or we should update the docs to mark them optional)

paulbellamy (changes requested):

src/server.ts
paulbellamy: Is that reflected in soroban-rpc as well?

paulbellamy (approved):

PR #62, opened 2023-03-29
Update types to conform to the new diagnostic event fields.

### LINKED ISSUE #59: add support for smart contract diagnostic events
### What
add support for smart contract debug events

### How

As a result, the code calling soroban-rpc's simulateTransaction and getEvents should be updated to match the changes at https://github.com/stellar/soroban-tools/issues/473 and https://github.com/stellar/soroban-tools/issues/475

TODO : we need a bit deeper analysis so we can better estimate time efforts. example : how many objects would be changed, which unit tests are going to be affected.

### Epic
https://github.com/stellar/soroban-tools/issues/471

Shaptic:
This adds the `diagnostic` event type and the `[]events` field to the `simulateTransaction()` response.

Closes #59.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "soroban-client",
-  "version": "0.4.1",
+  "version": "0.5.0",
   "description": "soroban-client is a library for working with the soroban-rpc server.",
   "main": "./lib/index.js",
   "types": "./lib/index.d.ts",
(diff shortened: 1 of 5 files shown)


paulbellamy (approved):
I was thinking that we should add a string enum type (`"system" | "contract" | "diagnostic"`) for the event filters. As a form of documentation. But LGTM either way.

PR #67, opened 2023-04-19
add server method for rpc getLatestLedger endpoint

sreuland:
Added new server method to wrap the rpc `getLatestLedger` endpoint.

Need access to this endpoint to obtain recent ledger number for polling contract events in a related dapp story - https://github.com/stellar/soroban-tools/issues/458

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -296,6 +296,24 @@ export class Server {
   }
 
   /**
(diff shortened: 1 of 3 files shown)


Shaptic (approved):
nice, thanks :clap:

PR #68, opened 2023-04-20
Remove phantomjs from dev dependencies

sreuland:
phantomjs is approaching eol and not published as a package for recent distros like ubuntu 22.04, it's not referenced by any tests here so, removing it as this is causing an issue in npm when building js-soroban-client from hosts that can't get the phantom js bin installed which is required on host path for `npm install` to work with this dependency in place.


paulbellamy (approved):
🚀

PR #69, opened 2023-04-20
SDK Modernization

Shaptic:
In line with stellar/js-stellar-sdk#814, this refreshes the build system to use the latest&greatest in JS build pipelines. 

It also promotes this library to `v1.0.0-beta.0`, which it may or may not be ready for.


chadoh (approved):
Huzzah!

sreuland (approved):
nice work, a veritable clinic, my approval is lightweight, look forward to porting this to `soroban` branch, you mentioned it resolves the webpack issue on js-xdr dependency from `stellar-base:9.0.0-soroban.1`, thanks!

quietbits (approved):

chadoh:
Seems like soroban-client no longer ships with TypeScript types. Is this change the cause?

chadoh:
Why use babel at all? Why not just build with TypeScript?

Shaptic:
that was a bug; #86 should have fixed it (part of v0.7.0)

as for Babel, I think we still need it to be able to control the output targets?

chadoh:
You mean this? https://www.typescriptlang.org/tsconfig#target

I'll try 0.7.0, especially now that futurenet has been updated! Thanks.

Shaptic:
Yeah so that one is nice but it's not granular enough, I think. Babel lets you target specific browser versions, market share, etc.

Actually, thinking about it more, we could probably get away with just using `tsc` here tbh. I need to figure out whether or not `ES6` matches up with Babel's `> 2%` as I outlined in #90, and then we can drop it (would be sweet).

PR #70, opened 2023-04-25
remove params from jsonrpc post payload if empty

sreuland:
when the jsonrpc.post() method:

```
export async function post<T>(
  url: string,
  method: string,
  ...params: any
)
```

 is webpacked for browser usage, there's some additional polyfill added to handle the spread `params` :

```
function post(url, method) {
    var _a;
    var params = [];
    for (var _i = 2; _i < arguments.length; _i++) {
        params[_i - 2] = arguments[_i];
    }

```

however, for methods such as getNetwork, getHealth, getLatestLedger that pass no `params` the result would still include `params: []` in the json rpc post, which rpc server rejects with err:

```
{
    "jsonrpc": "2.0",
    "id": 1,
    "error": {
        "code": -32602,
        "message": "no parameters accepted"
    }
}

```

https://github.com/stellar/soroban-tools/issues/458 will depend on this, it uses `getLatestLedger`

### DIFF
--- a/src/jsonrpc.ts
+++ b/src/jsonrpc.ts
@@ -35,6 +35,9 @@ export async function post<T>(
   method: string,
   ...params: any
 ): Promise<T> {
+  if (params && params.length < 1) {
+    params = null;
+  }
   const response = await axios.post<Response<T>>(url, {
     jsonrpc: "2.0",
     // TODO: Generate a unique request id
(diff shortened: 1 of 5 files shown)


tsachiherman (approved):
I think that this PR is good on it's own. However, I'm not sure if the server behavior is correct. Could you please create a ticket to check the soroban-rpc handling in these cases ? ( i.e. we need to align to spec, whatever that would be )

paulbellamy:
Yeah, seems like a bug in the server.

paulbellamy (approved):

PR #77, opened 2023-05-09
Use the new Soroban RPC simulation method when preparing transactions

### LINKED ISSUE #76: js-soroban-client: add support for fees
### What ?

txnbuilder and prepareTransaction need changes to support the fee model, because they need to set which resources the txn will use

### Epic

https://github.com/stellar/soroban-tools/issues/519

sreuland:
Adding support to 'prepare transaction' to apply new rpc simulation data, including multi-function auths, and suggested fees to the tx. Wanted to put the PR up asap get visibility and any course correction here if needed. working on unit tests now.

This depends on the rpc simulation changes - https://github.com/stellar/soroban-tools/pull/636/files

Closes #76


package.json
```diff
@@ -138,7 +138,7 @@
     "axios": "0.25.0",
     "es6-promise": "^4.2.4",
     "lodash": "4.17.21",
-    "stellar-base": "8.2.2-soroban.12",
+    "stellar-base": "https://github.com/stellar/js-stellar-base\\#host_functions_ts",
```
sreuland, line 141: I'll change this to `v9.0.0-soroban.1` when it's out, for now using this slightly newer stellar-base that has tweaks on txbuilder - [js-stellar-base#604](https://github.com/stellar/js-stellar-base/pull/604)

sreuland:
I'm working on fixing unit test failures.

src/transaction.ts
```diff
@@ -6,35 +6,40 @@ import {
...
     );
   }
 
   // TODO: Figure out a cleaner way to clone this transaction.
   const source = new Account(raw.source, `${parseInt(raw.sequence, 10) - 1}`);
-  const txn = new TransactionBuilder(source, {
-    fee: raw.fee,
+  const txnBuilder = new TransactionBuilder(source, {
+    fee: (
+      parseInt(raw.fee, 10) ||
```
sreuland, line 39: not sure on the algorithm for applying tx fees, we get a single `minResourceFee` from simulation which is a roll-up of all suggested resource fees, and it looked like that is to be treated additively to the existing classic fee, so just taking this simple approach to start, but need insights if this should change.

tsachiherman: The resource fees aren't "optional". It's guaranteed to fail if we don't meet these.
I'd suggest:
```js
max(0 + parseInt(simulation.minResourceFee, 10), 0 + parseInt(raw.fee, 10)).toString()
```

sreuland: @tsachiherman , I see the [inclusion fee requirement commented on preflight pr](https://github.com/stellar/soroban-tools/pull/636#discussion_r1190088092), how should we best integrate notion of inclusion fee in downstream UX, I see a couple variations on options:

1. soroban-client retains same tx helper interface, the rpc simulation response includes `suggestedInclusionFee`, which can be applied automatically within `assembleTransaction()` as final tx.fee = max(suggestedInclusionFee + minResourceFee, tx.fee), callers check resulting tx.fee to decide if they want to submit or handle otherwise. 

2. soroban-client retains same interface, sets final tx.fee to max(minResourceFee, tx.fee) and we document that callers must add a minimum additional inclusionFee of x stroops to the tx.fee before submit

3. soroban-client interfaces are changed to allow callers to pass `inclusionFee` on `assembleTransaction()` and `prepareTransaction()` and final tx.fee is set to max(inclusionFee + minResourceFee, tx.fee), callers check resulting tx.fee to decide whether to submit or handle otherwise.

tsachiherman: I think that #1 is the best option, since it would end up with the developer already having a good fees being placed within the transaction. The developer can always change that later on, but out-of-the-box, it would "just work".

src/transaction.ts
```diff
@@ -44,35 +63,47 @@ export function assembleTransaction(
...
+
+  return txnBuilder.build();
+}
+
+function buildExt(sorobanTxDataStr: string) {
+  const sorobanTxData: xdr.SorobanTransactionData = xdr.SorobanTransactionData.fromXDR(
+    sorobanTxDataStr,
+    "base64",
+  );
+  const txExt: xdr.TransactionExt = new xdr.TransactionExt(1);
```
sreuland, line 98: @Shaptic , do you have any insight on how to instantiate this TransactionExt union, i get a ts error on this constructor:

```
src/transaction.ts:98:60 - error TS2554: Expected 0 arguments, but got 1.

98   const txExt: xdr.TransactionExt = new xdr.TransactionExt(1);
```

it only happens from js-soroban-client, the same usage is working from [js-stellar-base in tx builder](https://github.com/stellar/js-stellar-base/pull/604/files#diff-38fee31808bd7d3eecb095e6b117fb4ea6c79843133da57fa7772f20de1fce6dR539), the one difference I can see is that the xdr is locally gen'd there rather than exported and then imported here. the exported version of it seems to be losing aspects of class construction.

src/transaction.ts
sreuland: ok, I've added new `suggestedInclusionFee` to the [rpc simulation response model here](https://github.com/stellar/js-soroban-client/pull/77/files#diff-718319220445fb0cf5b6fe5c0d0c88085f8f32ee0a90d41b496037b00836b8c3R124) and applying sum with `minResourceFee` to tx.fee automatically if is greater.

[mentioned need for this new attribute on rpc server side pr for preflight changes](https://github.com/stellar/soroban-tools/pull/636/files#r1190606646)

src/transaction.ts
```diff
@@ -44,35 +63,47 @@ export function assembleTransaction(
...
+    hostFnIndex < simulation.results.length;
+    hostFnIndex++
+  ) {
+    const rawHostFunction: xdr.HostFunction =
+      rawInvokeHostFunctionOp.functions[hostFnIndex];
+    const simHostFunctionResult: SorobanRpc.SimulateHostFunctionResult =
+      simulation.results[hostFnIndex];
+    rawHostFunction.auth(buildContractAuth(simHostFunctionResult.auth));
+    authDecoratedHostFunctions.push(rawHostFunction);
+  }
```
Shaptic, line 79: a more idiomatic approach would be, I think:

```suggestion
  // apply the pre-built Auth from simulation onto each Tx/Op/HostFunction 
  // invocation
  const authDecoratedHostFunctions = simulation.results.map((simResult, i) => {
    const hostFn: xdr.HostFunction = rawInvokeHostFunctionOp.functions[i];
    hostFn.auth(buildContractAuth(simResult.auth));
    return hostFn;
  });
```

src/transaction.ts
Shaptic: Oh jeez, I think the generator is busted on this? Can you do `xdr.TransactionExt[1]()`?

There's this in the Typescript:

```typescript
  class TransactionExt {
    switch(): number;

    sorobanData(value?: SorobanTransactionData): SorobanTransactionData;

    static 0(): TransactionExt;

    static 1(value: SorobanTransactionData): TransactionExt;

```

which is crazy: I don't think you can have only digits as a method name???

yarn.lock
```diff
@@ -1544,10 +1544,10 @@ big.js@^5.2.2:
   resolved "https://registry.yarnpkg.com/big.js/-/big.js-5.2.2.tgz#65f0af382f578bcdc742bd9c281e9cb2d7768328"
   integrity sha512-vyL2OymJxmarO8gxMr0mhChsO9QGwhynfuu4+MHTAW6czfq9humCB7rKpUjDd9YUiDPU4mzpyupFSvOClAwbmQ==
 
-bignumber.js@^4.0.0:
-  version "4.1.0"
-  resolved "https://registry.yarnpkg.com/bignumber.js/-/bignumber.js-4.1.0.tgz#db6f14067c140bd46624815a7916c92d9b6c24b1"
-  integrity sha512-eJzYkFYy9L4JzXsbymsFn3p54D+llV27oTQ+ziJG7WFRheJcNZilgVXMG0LoZtlQSKBsJdWtLFqOD0u+U0jZKA==
+bignumber.js@^9.1.1:
+  version "9.1.1"
```
Shaptic, line 1548: wait, where is this change coming from?

Shaptic: ah nvm it's from upgrading the base, right?

sreuland: yes, I changed the ref to stellar-base/host_functions_ts, and will repoint it at the new stellar-base release soon.

src/transaction.ts
sreuland: yes, can't instantiate the class by anything other than no-args or single number, one indicator here related to ts is that getting the compile error on constructor method args, I can comment that away with @ts-ignore, and then it constructs, but the object is not good, if i call the `sorobanData(xdr.SorobanTransactionData)` method on it, it takes it but it doesn't update the union's arm value, subsequent calls to `.value()` return undefined, which is where the problem finally stops as can't marshal the tx to xdr now due to invalid object, getting `Error: XDR Write Error: undefined is not a SorobanTransactionData`.

I added transactionbuilder tests in the stellar-base pr to test this same behavior on the gen xdr but from js test:
https://github.com/stellar/js-stellar-base/pull/604/files#diff-2f0f2abfad28c352b947215436ec683c35dc35dbe4854a3a79f2b4be522779e5R70

and am seeing the same result, the constructed object doesn't work, and those new tests are failing with same error when they try to marshal back to xdr.

src/transaction.ts
sreuland: oh nice, I didn't know you could get the the index also, yeah I had started with forEach first to do more functional, but still had a var outside holding index which made it clumsy, this is much cleaner, thanks for the pro tip!

src/transaction.ts
sreuland: @Shaptic , I think have narrowed the problem to js-xdr generated js for xdr unions with switches on number because the generated js for xdr unions with switch on types instead has correct js static factory methods based on the switch, and the objects generated by those work, I tested with:
```
let test = StellarBase.xdr.FeeBumpTransactionInnerTx.envelopeTypeTx(transaction.toEnvelope());
console.log("test value" + test.value());
```
the value returned matches the one passed on the static factory method, whereas for the gen'd code on number based switched unions, there is only that un-usable static factory method named with number, and using the default empty constructor does not work.

poking at js-sdk now ..

Shaptic: Thanks for the bug report! We'll have to use the parsing workaround for now, since that generator fix might take some time (legacy codebase, nobody _really_ knows how it works :facepalm:)

src/transaction.ts
```diff
@@ -6,35 +6,53 @@ import {
...
+
+  const rawInvokeHostFunctionOp: any = raw.operations[0];
+
+  if (
+    !rawInvokeHostFunctionOp.functions ||
+    !simulation.results ||
+    rawInvokeHostFunctionOp.functions.length !== simulation.results.length
+  ) {
+    throw new Error(
+      "preflight simulation results do not contain same count of HostFunctions that InvokeHostFunctionOp in the transaction has.",
```
Shaptic, line 43: I don't understand the intricacies of the multi-op host functions yet: why is the length enforced to be 1 above, but that isn't enforced here? I guess it's implicit, but I just want to make sure.

Related: is there no way to simulate transactions with multiple `invokeHostOp`erations?

src/transaction.ts
```diff
@@ -44,35 +62,47 @@ export function assembleTransaction(
...
+    "base64",
+  );
+
+  // TODO - remove this workaround to use constructor on js-xdr union
+  //       and use the typescript generated static factory method once fixed
+  //       https://github.com/stellar/dts-xdr/issues/5
+  // @ts-ignore
+  const txExt = new xdr.TransactionExt(1, sorobanTxData);
+  return txExt;
+}
```
Shaptic, line 100: related to my comment here: https://github.com/stellar/js-stellar-base/pull/604#discussion_r1194273887, this is what I meant about a more ergonomic API. Here, you would just do `txnBuilder.setSorobanData(simulation.transactionData)` and not worry about the extension/xdr/base64 nonsense.

src/transaction.ts
```diff
@@ -44,35 +62,47 @@ export function assembleTransaction(
...
+  );
+
+  // apply the pre-built Soroban Tx Ext from simulation onto the Tx
+  txnBuilder.setExt(buildExt(simulation.transactionData));
+
+  return txnBuilder.build();
+}
+
+function buildExt(sorobanTxDataStr: string) {
+  const sorobanTxData: xdr.SorobanTransactionData = xdr.SorobanTransactionData.fromXDR(
```
Shaptic, line 89: this type is pretty strongly implied by the static method already :wink:

src/transaction.ts
```diff
@@ -6,35 +6,53 @@ import {
...
-  const txn = new TransactionBuilder(source, {
-    fee: raw.fee,
+  const classicFeeNum = parseInt(raw.fee, 10) || 0;
+  const minResourceFeeNum = parseInt(simulation.minResourceFee, 10) || 0;
+  const txnBuilder = new TransactionBuilder(source, {
+    // automatically update the 'classic' tx fee if min resource fees from simulation response
+    // surpass that value.
+    // 'classic' fees are measured as the product of tx.fee * 'number of operations', In soroban contract tx,
+    // there can only be single operation in the tx, so can safely make simplification of classic fee = tx.fee.
+    fee: Math.max(classicFeeNum + minResourceFeeNum, classicFeeNum).toString(),
```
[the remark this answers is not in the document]
tsachiherman: I think that the second argument to the `Math.max` function was supposed to be `raw.fee`

src/transaction.ts
```diff
@@ -6,35 +6,53 @@ import {
...
-  // TODO: Figure out a cleaner way to clone this transaction.
   const source = new Account(raw.source, `${parseInt(raw.sequence, 10) - 1}`);
-  const txn = new TransactionBuilder(source, {
-    fee: raw.fee,
+  const classicFeeNum = parseInt(raw.fee, 10) || 0;
+  const minResourceFeeNum = parseInt(simulation.minResourceFee, 10) || 0;
+  const txnBuilder = new TransactionBuilder(source, {
+    // automatically update the 'classic' tx fee if min resource fees from simulation response
+    // surpass that value.
+    // 'classic' fees are measured as the product of tx.fee * 'number of operations', In soroban contract tx,
```
tsachiherman, line 53: classic fees are `classicFeeNum * operations num` right ?

src/transaction.ts
sreuland: @dmkozh , can you confirm with the latest changes for multi-function InvokeHostFunctionOp, the limitation for a single InvokeHostFunctionOp in Tx is still being enforced? I thought have seen that mentioned in chat and was assuming that limitation on validation of tx here.

src/transaction.ts
sreuland: yes, on classic tx's that would be the way classic fees would result, with the limitation of only one operation in the soroban tx, have reduced to just classic fees = classicFeeNum(the parsed numeric value of user specified tx.fee).

src/transaction.ts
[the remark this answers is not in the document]
sreuland: ah, good point, max is un-needed at this point, the incoming tx.fee has really become the 'suggested inclusion fee' when using `server.prepareTransaction(tx, networkPassphrase)`.

I think we can only append minResourceFee to what user provides on incoming tx.fee and return the sum as the updated tx.fee on the returned transaction from `server.prepareTransaction(tx, networkPassphrase)`.  I've updated the js docs on that method to illustrate further.

src/transaction.ts
sreuland: yes, I can refactor that, it would encapsulate the xdr.TransactionExt construction stuff to just one place in stellar-base tx-builder, if/when new TransactionExt union types later, the builder would need new method for each union type, but that is a fine trade-off for having the stronger types up front on the builder to get that more fluent client usage.

sreuland: refactored builder, using `setSorobanData()` - https://github.com/stellar/js-stellar-base/pull/604/commits/177b9e9acb19274b22192a9b0720c32099b19f1d

src/transaction.ts
sreuland: this has course corrected a bit, landed with just summing tx.fee and simulation `minResourceFee`

https://github.com/stellar/js-soroban-client/pull/77#discussion_r1194296683

src/transaction.ts
Shaptic: > the builder would need new method for each union type, but that is a fine trade-off for having the stronger types up front on the builder to get that more fluent client usage

yep, totally agreed here, and hopefully we can have clean abstractions for future extensions, too, if need be

src/transaction.ts
Shaptic: ohhhh yeah sorry, you're right :facepalm: one op per tx, but the op can have many fns. got it! 

:laughing: too many layers :cake:

Shaptic (approved):
really nice test coverage :clap:

package.json
Shaptic: it's out! https://www.npmjs.com/package/stellar-base/v/9.0.0-soroban.1

PR #78, opened 2023-05-17
fix: move type libs to devDeps

chadoh:
This PR moves the type libraries to devDependencies, as they are only needed for development and not for production.


Shaptic:
heh thanks :muscle: this is superseded by #69 now, though

PR #89, opened 2023-05-23
fixed parsing on simulation response for non-auth

sreuland:
if no auth is needed for contract, then simulation response will have `auth:null`, adjust client parsing to match this.

I ran `yarn fmt` as well and it triggered some minor changes here.

### DIFF
--- a/src/transaction.ts
+++ b/src/transaction.ts
@@ -94,8 +94,12 @@ export function assembleTransaction(
 
 function buildContractAuth(auths: string[]): xdr.ContractAuth[] {
   const contractAuths: xdr.ContractAuth[] = [];
-  for (const authStr of auths) {
-    contractAuths.push(xdr.ContractAuth.fromXDR(authStr, "base64"));
+  
+  if (auths) {
+    for (const authStr of auths) {
+      contractAuths.push(xdr.ContractAuth.fromXDR(authStr, "base64"));
+    }
   }
+
   return contractAuths;
 }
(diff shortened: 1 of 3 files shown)


src/transaction.ts
```diff
@@ -94,8 +94,12 @@ export function assembleTransaction(
...
-  for (const authStr of auths) {
-    contractAuths.push(xdr.ContractAuth.fromXDR(authStr, "base64"));
+  
+  if (auths) {
+    for (const authStr of auths) {
+      contractAuths.push(xdr.ContractAuth.fromXDR(authStr, "base64"));
+    }
   }
+
   return contractAuths;
```
Shaptic, line 104: This whole thing can be one line, I think:

```suggestion
function buildContractAuth(auths?: string[]): xdr.ContractAuth[] {
  return auths?.map((auth) => xdr.ContractAuth.fromXDR(auth, "base64"));
}
```

test/unit/transaction_test.js
```diff
@@ -147,11 +134,34 @@ describe("assembleTransaction", () => {
...
             .address()
             .accountId()
-            .ed25519(),
-        ),
+            .ed25519()
+        )
       ).to.equal("GBZXN7PIRZGNMHGA7MUUUF4GWPY5AYPV6LY4UV2GL6VJGIQRXFDNMADI");
     });
 
+    it("simulate ignores non auth from simulation", () => {
```
Shaptic, line 142: nice coverage :clap:

Shaptic (approved):
just one small suggestion on idiomatic js, but otherwise :+1:

src/transaction.ts
Shaptic: Oh hm, I'm not sure how this behaves in the `null` case. I think a `?? []` at the end would fix it.

sreuland: almost, but it changes the return type to `ContractAuth[] | undefined` which wasn't the intent, want to adapt whatever simulation data was present into an array instance, empty or otherwise.

PR #95, opened 2023-05-25
Release v0.7.1, fixing browser bundling bugs.

Shaptic:
This partially fixes #94, allowing the library to be imported as:

```ts
import * as SorobanClient from 'soroban-client';
```

in browser environments.

It also broadens the bundles' JS compatibility, reverting #90.

PR #96, opened 2023-05-25
Prepare v0.7.2 for release, further increasing downstream JS compatibility.

Shaptic:
After v0.7.1 (see #95), we broadened compatibility by transpiling browser bundles to an older JS feature set.

Unfortunately, this was insufficient, as Babel ignores any dependencies in Webpack's transpilation step if using a `.babelrc` file. By migrating to a `babel.config.json` file (see [docs here](https://babeljs.io/docs/configuration#whats-your-use-case), "You want to compile `node_modules`?") and adding the appropriate dependencies to the webpack config (note the delta on the `exclude` line), we can transpile these to the same feature set as the library itself.

cc @esteblock if you want to include this in your release, but it isn't required :+1:

### DIFF
--- a/webpack.config.browser.js
+++ b/webpack.config.browser.js
@@ -35,11 +35,11 @@ const config = {
     rules: [
       {
         test: /\.m?(ts|js)$/,
-        exclude: /node_modules/,
+        exclude: /node_modules\/(?!(stellar-base|js-xdr))/,
         use: {
           loader: "babel-loader",
           options: {
-            cacheDirectory: true,
+            cacheDirectory: true
           },
         },
       },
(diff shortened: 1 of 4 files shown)


piyalbasu (approved):

webpack.config.browser.js
```diff
@@ -35,11 +35,11 @@ const config = {
     rules: [
       {
         test: /\.m?(ts|js)$/,
-        exclude: /node_modules/,
+        exclude: /node_modules\/(?!(stellar-base|js-xdr))/,
```
sreuland, line 38: is this related to soroban-client exporting these(base and xdr)?

Shaptic: correct, and it needs to transpile them because they use newer JS features

PR #97, opened 2023-05-31
feat: allow server.getContractData's contractId to be a Contract Address

willemneal:


Shaptic (approved):
Nice!

src/server.ts
```diff
@@ -142,17 +143,20 @@ export class Server {
...
    * @param {xdr.ScVal} key - The key of the contract data to load.
    * @returns {Promise<SorobanRpc.LedgerEntryResult>} Returns a promise to the {@link SorobanRpc.LedgerEntryResult} object with the current value.
    */
   public async getContractData(
     contractId: string,
     key: xdr.ScVal,
   ): Promise<SorobanRpc.LedgerEntryResult> {
+    const buf = contractId.startsWith("C")
+      ? Address.fromString(contractId).toBuffer()
+      : Buffer.from(contractId, "hex");
```
paulbellamy, line 156: It could also be a hex string starting with 'C'...

Shaptic: Nice catch, you’d need a length check? Hacky but imo we do this for one release then drop hex support?

paulbellamy: Or we could add https://github.com/stellar/js-stellar-base/pull/614, and use `new Contract(contractId).address().toBuffer()`, which would handle the appropriate parsing.

paulbellamy: This `Contract.address()` helper was just released in `stellar-base@v9.0.0-soroban.2`, so if we update the `package.json`, we can use that.

paulbellamy (approved):

PR #103, opened 2023-06-28
Update client to support Preview 10

Shaptic:
**NEEDS RPC CHANGES MERGED FIRST.**

Done:
- [x] `yarn upgrade-interactive --latest` on dev dependencies 
- [x] upgrade stellar-base to preview 10
- [x] add durability (`'persistent' | 'temporary'`) support to ledger keys
- [x] move to single-function invokes
- [x] move to single-result simulations
- [x] move to new auth xdr structure

Remaining:
- [x] merge to remove `as unknown` hack: stellar/js-stellar-base#635 
- [x] fix tests
- [x] changelog

The major change is to `getContractData()`, adding an optional parameter:

```typescript
  /**
   * @param {Durability} [durability] - The "durability keyspace" that this
   *    ledger key belongs to, which is either 'temporary' or 'persistent' (the
   *    default), see {@link Durability}.
   * 
   * @warning If the data entry in question is a 'temporary' entry, it's
   *    entirely possible that it has expired out of existence. Future versions
   *    of this client may provide abstractions to handle that.
   **/
```


src/transaction.ts
```diff
@@ -66,20 +61,17 @@ export function assembleTransaction(
...
-      functions: authDecoratedHostFunctions,
-    }),
+    Operation.invokeHostFunction({
+      func: invokeOp.func,
+      auth: (invokeOp.auth ?? []).concat(
+        simulation.results[0].auth?.map((a: string) =>
+          xdr.SorobanAuthorizationEntry.fromXDR(a, "base64")
+        ) ?? []
+      ),
+    } as unknown as OperationOptions.InvokeHostFunction /* until .1 is released */),
```
Shaptic, line 74: this is fixed by https://github.com/stellar/js-stellar-base/pull/635

src/server.ts
```diff
@@ -141,43 +147,107 @@ export class Server {
...
+   *    {@link SorobanRpc.LedgerEntryResult} object with the current value.
+   *
+   * @warning If the data entry in question is a 'temporary' entry, it's
+   *    entirely possible that it has expired out of existence. Future versions
+   *    of this client may provide abstractions to handle that.
    */
   public async getContractData(
     contractId: string,
     key: xdr.ScVal,
+    expirationType?: 'temporary' | 'persistent'
```
tomerweller, line 181: A few thoughts 
1. Should we remove the optionality from the storage type? This feels unnecessarily magical and we're introducing more than one way to do something very simple. When a developer queries storage directly it assumes a familiarity with the storage layout of the contract and so it's expected that they know which type of storage is used. 
2. Should we change the api to more closely resemble the SDK? Often times it's the same developer(s) writing the contract and dapps so similar is better (to the extent possible). In the SDK we made a conscious decision to have clear namespace separation for persistent and temporary storage. example `env.storage().persistent().get()`. Maybe in this case just have two separate functions?
3. On the naming front, a few sdk iterations ago we replaced `data` with `storage`. Should we use `storage` here too? to complicate things further the json-rpc method is called `getLedgerEntries` so we have the same thing called three different names across three different layers of the stack - what a party :)

tomerweller: > Should we change the api to more closely resemble the SDK

Probably no. Given that this is the `Server` object I think we should be as close as possible to the json-rpc API, not the SDK. In that case, should we change it to be `getLedgerEntries(keys)` similar to the server? We can introduce helpers elsewhere for creating keys.    

(Whether the sdk and rpc should be more similar is a different question I guess)

Shaptic: Yeah I'm certainly not opposed to durability being explicit (discussed in [Slack thread](https://stellarfoundation.slack.com/archives/C02B04RMK/p1687986170249409)). Two separate functions seems like overkill, but you make a good point about developers having experience transfer from contract to client. I changed the behavior in ec67a12, wdyt?

For your latter points, there already **is** a `Server.getLedgerEntries(keys: xdr.LedgerKey[])`, so there should not be much confusion. This method, in fact, is intended to be a helper oriented around fetching `LedgerKeyContractData`s, hence the name (over "storage") and attempt at keeping it easy to use.

tomerweller: oh cool. looks good and thanks for the clarification!

src/server.ts
```diff
@@ -135,49 +141,85 @@ export class Server {
...
+   *    {@link SorobanRpc.LedgerEntryResult} object with the current value.
+   *
+   * @warning If the data entry in question is a 'temporary' entry, it's
+   *    entirely possible that it has expired out of existence. Future versions
+   *    of this client may provide abstractions to handle that.
    */
   public async getContractData(
     contractId: string,
     key: xdr.ScVal,
+    expirationType: 'temporary' | 'persistent' = 'temporary'
```
paulbellamy, line 174: 3 suggestions:
- Let's call this durability to match the API elsewhere, and in the xdr.
- wdyt of making `persistent` the default? That is what we are pushing people towards.
- Maybe extract the Durability to a type or enum?
```suggestion
    durability: Durability = 'persistent'
```

src/server.ts
```diff
@@ -135,49 +141,85 @@ export class Server {
...
+   *
+   * @returns {Promise<SorobanRpc.LedgerEntryResult>} Returns a promise to the
+   *    {@link SorobanRpc.LedgerEntryResult} object with the current value.
+   *
+   * @warning If the data entry in question is a 'temporary' entry, it's
+   *    entirely possible that it has expired out of existence. Future versions
+   *    of this client may provide abstractions to handle that.
    */
   public async getContractData(
     contractId: string,
```
paulbellamy, line 172: Minor suggestion: Might be convenient if this also took a few more variants.
```suggestion
    contractId: string | Address | ScAddress,
```

src/server.ts
```diff
@@ -135,49 +141,85 @@ export class Server {
...
-      }),
+        durability,
+        bodyType: xdr.ContractEntryBodyType.dataEntry()   // expirationExtension is internal
+      })
     ).toXDR("base64");
-    const getLedgerEntriesResponse: SorobanRpc.GetLedgerEntriesResponse = await jsonrpc.post(
+
+    const getLedgerEntriesResponse = await jsonrpc.post<
+      SorobanRpc.GetLedgerEntriesResponse
+    >(
```
paulbellamy, line 201: Suggestion: Seems like we could simplify this with...
```ts
return this.getLedgerEntries([xdr.LedgerKey.contractData(...)])[0];
```

src/soroban_rpc.ts
```diff
@@ -131,7 +131,7 @@ export namespace SorobanRpc {
     transactionData: string;
     events: string[];
     minResourceFee: string;
-    results: SimulateHostFunctionResult[];
+    results: SimulateHostFunctionResult[]; // always one element, tho
```
paulbellamy, line 134:
```suggestion
    // This will only contain a single element, because only a single invokeHostFunctionOperation is supported per transaction.
    results: SimulateHostFunctionResult[];
```

paulbellamy (approved):

src/server.ts
Shaptic: Oh really? I thought temporary was actually preferred (i.e. if you want to pay for and worry about persistence, you should know its purpose & understand it better).

tomerweller: I think that this dialog suggests that not having a default might be the preferred way to go?

src/server.ts
Shaptic: good idea, added `Contract | Address`. I don't think `xdr.ScAddress` is useful? and also hard to detect because you can't `instanceof` it reliably, I think

src/server.ts
Shaptic: I think the original intent was a better error on missing keys? (I didn't write this helper in the first place.) I cleaned it up based on your suggestion though, e.g. no reason to `await` when returning a `Promise`, anyway.

src/server.ts
Shaptic: I guess it begs the question of "how much do we want people to care about storage durability?" is the answer "when it's important", "usually", or "always"?

paulbellamy (approved):

PR #107, opened 2023-07-03
Prepare v0.9.0 for release.

Shaptic:
# Preview 10 Release Notes

This Preview outlines a foundation for State Expiration on the Soroban platform.

## XDR Changes

Significant changes have happened in the way that authorization trees are defined for contract invocations. This **should** be covered by transaction simulation, but you can view stellar/js-stellar-base#633 and stellar/js-stellar-base#634 for details and diffs around the changes.

### Breaking Changes
- The `Server.getContractData` method now takes an optional parameter of a new type, `Durability`, representing the "keyspace" of the contract data. It should be set to either `Durability.Persistent` (the default) or `Durability.Temporary`, depending on the type of storage that backs this particular piece of data.
- The `Operation.invokeHostFunction` method now takes a `func` parameter that should be an `xdr.HostFunction` instead of an `args` parameter. 
- The `Operation.invokeHostFunctions` method has been removed because multiple host function invocations in a single operation are no longer supported.

### New Operations
- To facilitate bumping rent and expiration for contract ledger entries, there are two new operations:
 * `Operation.bumpFootprintExpiration({ ledgersToExpire: number })` bumps the expiration all of the read and written ledger keys in the transaction's `sorobanData` by the specified number of ledgers
 * `Operation.restoreFootprint()` uses the transaction's `sorobanData` field to restore the entire footprint
* `Server.prepareTransaction` now incorporates simulation results containing the new `BumpFootprintExpirationOp`s and `RestoreFootprintOp`s as part of `assembleTransaction`.

## New Features

We've made an effort to improve the abstractions around dealing with smart contract values (`xdr.ScVal`). There are a handful of new helpful interfaces to make dealing with them easier:

### Large Integers
"Large" integers (e.g. 64, 128, and 256-bit values) and their `ScVal` equivalents (for passing to raw `Operation.invokeHostOperation` structures, to `Contract.call(...)` invocations, or as part of the authorization framework) can now easily be crafted from `string`s or `bigint`s. 

**You should never need to deal with bitwise operations or endianness again!**

This is accomplished via the following APIs:

```ts
class ScInt {
  constructor(
    value: number | string | bigint | ScInt,
    opts?: { type: ScIntType }
  );

  toU64(): xdr.ScVal;
  /* ... and the others ... */

  toString(): string;
  toBigInt(): bigint;
  toNumber(): number;
}

function scValToBigInt(scv: xdr.ScVal): bigint;
```

The interfaces should be pretty self-explanatory, but you can refer to [the documentation](https://stellar.github.io/js-soroban-client/global.html#scValToBigInt) for details and example usage. To keep it simple, you can do:

```ts
let input = "1000000000000";
let scv = new ScInt(input).toU128();
```

Here, `scv` is an `xdr.ScVal` with the `U128` type. Similarly, you can get the integer back out:

```ts
let bigi = scValToBigInt(scv);
bigi === 1000000000000n
```

### Native Conversions

There are also new abstractions to easily convert between native JavaScript types and `xdr.ScVal`s. Since they are new and very high-level, they try to make reasonable assumptions about what you hope to accomplish.

**You should never need to deal with converting basic (and nested basic) types to and from XDR smart contract values (`ScVal`s)!**

The APIs are:

```ts
function nativeToScVal(n: any, opts?: { type: any }): xdr.ScVal;
function scValToNative(n: xdr.ScVal): any;
```

The [documentation](https://stellar.github.io/js-soroban-client/global.html#nativeToScVal) has details on each conversion and what types of conversions you can force via options, but here's some simple examples:

```ts
const native = {
  name: "soroban",
  age: 123,
  interests: [
    "smart", 
    "contract", 
    "abstractions"
  ]
}

const scv = nativeToScVal(native, { 
  type: {
    // force the age to be interpreted as a symbol key and i128 value
    age: ['symbol', 'i128']
    // all other entries will have default conversions
  }
});
// naturally, scValToNative(scv) == scv

// you can do the same type-interpretation as above for large integers
const scvAge = nativeToScVal(native.age, 'u32');
```

It handles most of the `ScVal` types you will see in high-level contexts and even handles nested types.

### DIFF
--- a/src/transaction.ts
+++ b/src/transaction.ts
@@ -77,6 +77,9 @@ export function assembleTransaction(
       break;
 
     case "bumpFootprintExpiration":
+      txnBuilder.addOperation(Operation.bumpFootprintExpiration(raw.operations[0]));
+      break;
+
     case "restoreFootprint":
       txnBuilder.addOperation(Operation.restoreFootprint(raw.operations[0]));
       break;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ A breaking change should be clearly marked in this log.
 ## Unreleased
 
 
+## v0.9.0
+
 ### Updated
 * `Server.getContractData` has an additional, optional parameter: `expirationType?: string` which should be set to either `'temporary'` or `'persistent'` depending on the type of ledger key. By default, it will attempt to fetch both, returning whichever one it finds ([#103](https://github.com/stellar/js-soroban-client/pull/103)).
 * `assembleTransaction` now accepts simulation results for the new `BumpFootprintExpirationOp`s and `RestoreFootprintOp`s ([#108](https://github.com/stellar/js-soroban-client/pull/108)).
--- a/package.json
+++ b/package.json
@@ -124,7 +124,7 @@
     "mocha": "^10.2.0",
     "node-polyfill-webpack-plugin": "^2.0.1",
     "nyc": "^15.1.0",
-    "prettier": "^3.0.0",
+    "prettier": "^2.8.8",
     "randombytes": "^2.1.0",
     "sinon": "^15.2.0",
     "sinon-chai": "^3.7.0",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "soroban-client",
-  "version": "0.8.1",
+  "version": "0.9.0",
   "description": "A library for working with Stellar's Soroban RPC servers.",
   "author": "Stellar Development Foundation <hello@stellar.org>",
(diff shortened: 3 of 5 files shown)


src/transaction.ts
```diff
@@ -77,6 +77,9 @@ export function assembleTransaction(
       break;
 
     case "bumpFootprintExpiration":
+      txnBuilder.addOperation(Operation.bumpFootprintExpiration(raw.operations[0]));
```
Shaptic, line 80: :facepalm: a mistake from #108 which is fixed by the new test in this PR

tsachiherman (approved):

PR #108, opened 2023-07-10
Add support for Preview 10 operations when assembling Soroban transactions.

Shaptic:
This builds on #107 to allow `BumpFootprintExpirationOp`s and `RestoreFootprintOp`s to be assembled when users call `simulateTransaction` and `prepareTransaction`.

### DIFF
--- a/src/transaction.ts
+++ b/src/transaction.ts
@@ -13,30 +13,28 @@ import { SorobanRpc } from "./soroban_rpc";
 export function assembleTransaction(
   raw: Transaction | FeeBumpTransaction,
   networkPassphrase: string,
-  simulation: SorobanRpc.SimulateTransactionResponse,
+  simulation: SorobanRpc.SimulateTransactionResponse
 ): Transaction {
   if ("innerTransaction" in raw) {
     // TODO: Handle feebump transactions
     return assembleTransaction(
       raw.innerTransaction,
       networkPassphrase,
-      simulation,
+      simulation
     );
   }
 
-  if (
-    raw.operations.length !== 1 ||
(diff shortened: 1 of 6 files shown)


src/transaction.ts
```diff
@@ -61,25 +59,63 @@ export function assembleTransaction(
...
-        ) ?? []
-      ),
-    })
-  );
+  switch (raw.operations[0].type) {
+    case "invokeHostFunction":
+      const invokeOp: Operation.InvokeHostFunction = raw.operations[0];
+      txnBuilder.addOperation(
+        Operation.invokeHostFunction({
+          source: invokeOp.source,
```
Shaptic, line 67: note that previously we weren't copying the source

bitcoinfinancier (approved):

src/transaction.ts
```diff
@@ -61,25 +59,63 @@ export function assembleTransaction(
...
+              xdr.SorobanAuthorizationEntry.fromXDR(a, "base64")
+            ) ?? []
+          ),
+        })
+      );
+      break;
+
+    case "bumpFootprintExpiration":
+      const bumpOp: Operation.BumpFootprintExpiration = raw.operations[0];
+      txnBuilder.addOperation(
```
sreuland, line 81: since this and restore case are not merging any data from simulation, can it reduce to `txnBuilder.addOperation(raw.operations[0])`, or I may be missing a subtlety.

sreuland (approved):

src/transaction.ts
Shaptic: I thought so, too, but kept getting some TypeScript errors... tried it again and it worked :facepalm: Great call, thank you!

PR #110, opened 2023-07-14
Handle null ledger entries and return a better error for missing accounts

paulbellamy:
This is the js client version of https://github.com/stellar/soroban-tools/pull/771

If the soroban-rpc server finds no matching entries for a `getLedgerEntries`, it will return `entries: null`. The client should handle this.

Also, if we try to do a `getAccount` and don't find it, we can return a more helpful error to the user.

### DIFF
--- a/src/soroban_rpc.ts
+++ b/src/soroban_rpc.ts
@@ -34,7 +34,7 @@ export namespace SorobanRpc {
   /* Response for jsonrpc method `getLedgerEntries`
    */
   export interface GetLedgerEntriesResponse {
-    entries: LedgerEntryResult[];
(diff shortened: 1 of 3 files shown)


src/soroban_rpc.ts
```diff
@@ -34,7 +34,7 @@ export namespace SorobanRpc {
   /* Response for jsonrpc method `getLedgerEntries`
    */
   export interface GetLedgerEntriesResponse {
-    entries: LedgerEntryResult[];
+    entries: LedgerEntryResult[] | null;
```
Shaptic, line 37: why couldn't it just be empty instead of null :sob:

Shaptic (approved):

PR #111, opened 2023-07-15
Fix bigIntFromBytes

lmorgan824:
Line 702 I think is testing the high bit of bytes[0], not the bytes[0] === 0x80
Changed to bytes[0] & 0x80


Shaptic:
Thanks for catching this! This code has no business being in here, anyway. I dropped it entirely in #112.

PR #114, opened 2023-07-26
Preserve existing auth entries after simulation if they're present

Shaptic:
In complex auth scenarios, you may have signed entries that would be overwritten by simulation, so this just uses your existing entries. This should not affect "new" or "fresh" simulations just like before.

### DIFF
--- a/src/transaction.ts
+++ b/src/transaction.ts
@@ -66,18 +66,19 @@ export function assembleTransaction(
         Operation.invokeHostFunction({
           source: invokeOp.source,
           func: invokeOp.func,
-          // apply the auth from the simulation
-          auth: (invokeOp.auth ?? []).concat(
-            simulation.results[0].auth?.map((a: string) =>
-              xdr.SorobanAuthorizationEntry.fromXDR(a, "base64")
-            ) ?? []
-          ),
+          // override auth from the simulation: any existing `invokeOp.auth`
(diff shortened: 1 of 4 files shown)


aristidesstaffieri (approved):
lgtm, thanks!

src/transaction.ts
```diff
@@ -62,22 +62,30 @@ export function assembleTransaction(
...
         Operation.invokeHostFunction({
           source: invokeOp.source,
           func: invokeOp.func,
-          // apply the auth from the simulation
-          auth: (invokeOp.auth ?? []).concat(
-            simulation.results[0].auth?.map((a: string) =>
-              xdr.SorobanAuthorizationEntry.fromXDR(a, "base64")
-            ) ?? []
-          ),
+          // if auth entries are already present, we consider this "advanced
```
sreuland, line 70: minor suggestion, this is really good description on the new behavior for auth based on incoming tx, which could be included in public facing js doc on [prepareTransaction()](https://github.com/stellar/js-soroban-client/pull/114/files#diff-8a8ae07582c9d433ec8c2e5c4310ff8901e604f4965c5b90a49117ad46c47595R466)

sreuland (approved):

src/transaction.ts
Shaptic: great idea! done in 12973ee

PR #115, opened 2023-07-27
feat:ContractSpec class to parse ScVals

willemneal:
This adds a new ContractSpec class which is comprised of the spec entries of a contract. This allows the internal user defined types to be accessible making the parsing to and from XDR more accurate. This should be moved to the js-stellar-base, but for now it is best served here for the type support.

This should replace the current parsing logic in the TS bindings currently living in the soroban-tools repo. 

Also discussion here: https://gist.github.com/tyvdh/9391482f7490b085b283ea96d2f50b11


test/unit/strval_test.js
```diff
@@ -0,0 +1,59 @@
...
+  });
+
+});
+
+
+function custom_types_spec() {
+  return SorobanClient.ContractSpec(xdr.Sp)
+}
+
+const CUSTOM_TYPES_XDR = ["AAAAAQAAAC9UaGlzIGlzIGZyb20gdGhlIHJ1c3QgZG9jIGFib3ZlIHRoZSBzdHJ1Y3QgVGVzdAAAAAAAAAAABFRlc3QAAAADAAAAAAAAAAFhAAAAAAAABAAAAAAAAAABYgAAAAAAAAEAAAAAAAAAAWMAAAAAAAAR",
```
paulbellamy, line 26: These might be easier to maintain if we have the entry as a JS object, and serialize it in the test. That would mean fewer opaque strings to maintain here if/when the XDR changes.

src/strval.ts
paulbellamy, line 1: I'd suggest naming this `src/contract_spec.ts`. I believe we normally name files after the "main class" in the file.

paulbellamy (commented):
A couple small suggestions, but looks awesome.

test/unit/strval_test.js
willemneal: Yeah this was me just trying to get a valid XDR string.

src/contract_spec.ts
```diff
@@ -0,0 +1,643 @@
...
+          case "scSpecTypeU128":
+            return new ScInt(val, { type: "u128" }).toU128();
+          case "scSpecTypeI128":
+            return new ScInt(val, { type: "i128" }).toI128();
+          case "scSpecTypeU256":
+            return new ScInt(val, { type: "u256" }).toU256();
+          case "scSpecTypeI256":
+            return new ScInt(val, { type: "i256" }).toI256();
+          default:
+            throw new TypeError(`invalid type (${ty}) specified for integer`);
```
Shaptic, line 314: suggestion to significantly simplify this: 
```typescript
// convert well-formed spec type to int type, e.g. 'scSpecTypeU256' --> 'u256'
const intType = ty.switch().name.substring(10).toLowerCase();
return new XdrLargeInt(intType, val as number).toScVal();
```

`XdrLargeInt` will do the error-handling, or you can try-catch if you want to bubble up `ty` specifically.

willemneal: Will do!

src/contract_spec.ts
```diff
@@ -0,0 +1,643 @@
...
+   *     'hello': [ 'symbol', 'i128' ],
+   *   }
+   * })
+   * // gives scvMap with entries: [
+   * //     [ scvSymbol, scvI128 ],
+   * //     [ scvString, scvArray<scvBool> ]
+   * // ]
+   * ```
+   */
+  nativeToScVal(val: any, typeDef?: xdr.ScSpecTypeDef): xdr.ScVal {
```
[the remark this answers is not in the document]
willemneal: Yeah my goal was to clean up the docs today.

src/index.ts
```diff
@@ -7,6 +7,7 @@ const version = require("../package.json").version;
 
 // Expose all types
 export * from "./soroban_rpc";
+export * from "./contract_spec";
```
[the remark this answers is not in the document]
willemneal: I had been using it to export some stuff for tests since it was much easier to construct the example ScSpecTypeDef types by hand. I can move it to the tests and switch to the singe output so it's a smaller API surface.

src/contract_spec.ts
willemneal: Also wouldn't I want `as bigint` ?

Shaptic: Bah, that's a typescript bug then - the js itself has a `toScVal()`: https://github.com/stellar/js-stellar-base/blob/soroban/src/numbers/xdr_large_int.js#L202, so you can `@ts-ignore` for now. As for `bigint`, I think you're right - I didn't read into that part closely and just borrowed the `as number` you had.

Shaptic: Bah, that's a typescript bug then - the js itself has a `toScVal()`: https://github.com/stellar/js-stellar-base/blob/soroban/src/numbers/xdr_large_int.js#L202, so you can `@ts-ignore` for now. As for `bigint`, I think you're right - I didn't read into that part closely and just borrowed the `as number` you had.

src/index.ts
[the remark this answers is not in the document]
Shaptic: Totally fair re: testing!

src/contract_spec.ts
```diff
@@ -0,0 +1,643 @@
...
+    }
+    let num = scv.value() as number;
+    if (udt.cases().some((entry) => entry.value() === num)) {
+    }
+    return num;
+  }
+}
+
+function stringToScVal(
+  str: string | Uint8Array,
```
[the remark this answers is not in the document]
Shaptic: bytearrays are already way too easy to add unnecessary complexity here, i.e. `xdr.ScVal.scvBytes(...)`

src/contract_spec.ts
Shaptic: This is fixed in the latest stellar-base!

src/contract_spec.ts
```diff
@@ -0,0 +1,722 @@
...
+
+  /**
+   * Constructs a new ContractSpec from an array of XDR spec entries.
+   *
+   * @param {xdr.ScSpecEntry[] | string[]} entries - The XDR spec entries
+   * 
+   * @throws {Error} If entries is invalid
+   */
+  constructor(entries: xdr.ScSpecEntry[] | string[]) {
+    if (entries.length > 0) {
```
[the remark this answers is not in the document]
willemneal: Good idea!

src/contract_spec.ts
```diff
@@ -0,0 +1,722 @@
...
+      throw new Error(`no such entry: ${name}`);
+    }
+    return entry;
+  }
+
+  /**
+   * Converts a native JS value to an ScVal based on the given type.
+   *
+   * @param {any} val - The native JS value
+   * @param {xdr.ScSpecTypeDef} [typeDef] - The expected type
```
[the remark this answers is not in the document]
willemneal: Very true! I'll clean it up!

willemneal:
@Shaptic @paulbellamy I removed having the optional typeDef and it simplified everything. I was also able to switch all string comparisons to value comparisons which should improve performance (though I'm sure that's not the bottleneck currently for devs).

willemneal:
@paulbellamy I went with still having the base64 as it was easiest to copy over from the soroban-cli tests we added for the bindgen.  For now I can regenerate the json file manually, but I'll add a script to do it soon.

test/unit/contract_spec.ts
```diff
@@ -0,0 +1,277 @@
+import { xdr, ContractSpec, Address } from "../../lib";
+//@ts-ignore
+import spec from "../spec.json";
+import { expect } from "chai";
+
+const publicKey = "GCBVOLOM32I7OD5TWZQCIXCXML3TK56MDY7ZMTAILIBQHHKPCVU42XYW";
+const addr = Address.fromString(publicKey);
+
+describe("Can parse base64 entries", function () {
+  const SPEC = new ContractSpec(spec);
```
Shaptic, line 10: this should be in an `it`, I think, to catch errors? like so:

```suggestion
let SPEC;

it('constructs from base64 entries', function() {
  SPEC = new ContractSpec(spec);
}

it('throws with no spec', function() {
  expect(() => new ContractSpec([])).to.throw(/expected error message here/i);
}
```

test/unit/contract_spec.ts
```diff
@@ -0,0 +1,277 @@
...
+
+  it("u128", () => {
+    roundtrip("u128", 1n);
+  });
+
+  it("map", () => {
+    const map = new Map();
+    map.set(1, true);
+    map.set(2, false);
+    // map.set(3, 'hahaha') // should throw an error
```
Shaptic, line 145: can we test this?
```suggestion
    expect(() => map.set(3, 'hahaha')).to.throw(/what error?/i);
```

test/unit/contract_spec.ts
Shaptic: whereas the `describe` should be something more like "build values using ContractSpec"

package.json
```diff
@@ -29,12 +29,13 @@
     "build:prod": "cross-env NODE_ENV=production yarn _build",
     "build:node": "yarn _babel && yarn build:ts",
     "build:ts": "tsc -p ./tsconfig.json",
+    "build:test": "yarn tsc -p ./test/tsconfig.json",
```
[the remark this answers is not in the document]
Shaptic: also, either way, you don't need the `yarn` here

test/unit/contract_spec.ts
```diff
@@ -0,0 +1,277 @@
...
+  });
+});
+
+describe("parsing and building ScVals", function () {
+  it("Can parse entries", function () {
+    let spec = new ContractSpec([GIGA_MAP, func]);
+    let fn = spec.findEntry("giga_map");
+    let gigaMap = spec.findEntry("GigaMap");
+    expect(gigaMap).deep.equal(GIGA_MAP);
+    expect(fn).deep.equal(func);
```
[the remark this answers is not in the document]
willemneal: Good to know!

package.json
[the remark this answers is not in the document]
willemneal: good point! I'm just used to doing that to ensure the local compiler.

I didn't want the tests to have the same config as the main one. Also currently have a limitation in that I needed the test files to generate JS in place. I couldn't figure out how to have the relative imports work otherwise.

test/unit/contract_spec.ts
willemneal: JS doesn't care if the values have conflicting types. So I added a test that tests the error you get when trying to do the round trip again after adding the string.

willemneal:
@Shaptic Updated after your comments. Definitely could test more errors, but I think the tests are good enough for now

Shaptic (approved):
Really **really** awesome work! Excited to get this across the finish line.

Last question: this uses futurenet-compatible XDR, right? I think yes based on the package.json but I want to confirm.

willemneal:
Yep! That is why I ported the CLI's tests! But using this to do the parsing of the generated bindings I was able to get a roundtrip with futurenet (we don't have the tests in CI in that repo yet because we don't want to depend on CI).

Shaptic:
Love it! So are we ready to merge, then? It'd be awesome if you could add a brief changelog entry that matches the format of the others, then I can hit the button :shipit:

PR #117, opened 2023-08-04
Explicitly drops support for hex-encoded contract IDs.

Shaptic:
This is related to stellar/js-stellar-base#650, which fully drops support for the now-deprecated hex encoding method for contracts. The [SEP-23 strkey format](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0023.md#specification) is mandatory, now. You can migrate as such:

```typescript
const strkey = StrKey.encodeContract(Buffer.from(hexContractId, 'hex'));
```

To enforce this, we also upgraded `stellar-base`.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -401,7 +401,7 @@ export class Server {
    * ledger footprint, expected authorizations, and expected costs.
    *
    * @example
-   * const contractId = '0000000000000000000000000000000000000000000000000000000000000001';
+   * const contractId = 'CA3D5KRYM6CB7OWQ6TWYRR3Z4T7GNZLKERYNZGGA5SOAOPIFY6YQGAXE';
    * const contract = new SorobanClient.Contract(contractId);
    *
    * // Right now, this is just the default fee for this example.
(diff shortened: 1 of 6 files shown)


src/server.ts
```diff
@@ -242,9 +242,9 @@ export class Server {
    * To fetch contract wasm byte-code, use the ContractCode ledger entry key.
    *
    * @example
-   * const contractId = "0000000000000000000000000000000000000000000000000000000000000001";
+   * const contractId = "CAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAD2KM";
```
sreuland, line 245: minor, I noticed similar examples with the hex id string usage in js-docs of other methods in server.ts like[ getLedgerEntries() ](https://github.com/stellar/js-soroban-client/blob/421d2bf5ef0fd8b0ff08b2508e4ceb21693d485c/src/server.ts#L404-L405), is it worthwhile for consistency to give those the same update?

sreuland (approved):

src/server.ts
Shaptic: Thought I got em all but my `Ctrl+F`-fu wasn't great. Thanks for pointing this out! Should've got 'em all, now.

PR #120, opened 2023-08-07
Simplify transaction assembly after new base features

Shaptic:
**This releases v0.10.1.** 

The changelog is just this PR:

### Fixed
* The `stellar-base` dependency has been upgraded to fix a TypeScript bug ([#665](https://github.com/stellar/js-stellar-base/pull/665)).
* Bundle size has decreased by refactoring `assembleTransaction` to use new abstractions from `stellar-base` ([#120](https://github.com/stellar/js-soroban-client/pull/120)).

### DIFF
--- a/src/transaction.ts
+++ b/src/transaction.ts
@@ -1,5 +1,4 @@
 import {
-  Account,
   FeeBumpTransaction,
   Operation,
   Transaction,
@@ -9,7 +8,23 @@ import {
 
 import { SorobanRpc } from "./soroban_rpc";
 
-// TODO: Transaction is immutable, so we need to re-build it here. :(
+/**
+ * Combines the given raw transaction alongside the simulation results.
+ *
+ * @param raw   the initial transaction, w/o simulation applied
+ * @param networkPassphrase   the network this simulation applies to
+ * @param simulation  the Soroban RPC simulation result
+ *
+ * @returns a new, cloned transaction with the proper auth and resource (fee,
+ *    footprint) simulation data applied
+ *
+ * @note if the given transaction already has authorization entries in a host
+ *    function invocation (see {@link Operation.invokeHostFunction}), **the
+ *    simulation entries are ignored**.
+ *
+ * @see {Server.simulateTransaction}
+ * @see {Server.prepareTransaction}
+ */
 export function assembleTransaction(
   raw: Transaction | FeeBumpTransaction,
   networkPassphrase: string,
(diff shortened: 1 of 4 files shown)


tsachiherman (approved):

aristidesstaffieri (approved):

src/transaction.ts
```diff
@@ -49,18 +64,16 @@ export function assembleTransaction(
...
-    minAccountSequenceAge: raw.minAccountSequenceAge,
-    minAccountSequenceLedgerGap: raw.minAccountSequenceLedgerGap,
-    extraSigners: raw.extraSigners,
+    // apply the pre-built Soroban Tx Data from simulation onto the Tx
+    sorobanData: simulation.transactionData,
   });
 
   switch (raw.operations[0].type) {
     case "invokeHostFunction":
+      // @ts-ignore
```
sreuland, line 73: not looking through an IDE atm, what was the warning to ignore?

src/transaction.ts
```diff
@@ -36,10 +51,10 @@ export function assembleTransaction(
     throw new Error(`simulation results invalid: ${simulation.results}`);
   }
 
-  const source = new Account(raw.source, `${parseInt(raw.sequence, 10) - 1}`);
   const classicFeeNum = parseInt(raw.fee, 10) || 0;
   const minResourceFeeNum = parseInt(simulation.minResourceFee, 10) || 0;
-  const txnBuilder = new TransactionBuilder(source, {
+
+  const txnBuilder = TransactionBuilder.cloneFrom(raw, {
```
sreuland, line 57: nice!

sreuland (approved):

src/transaction.ts
Shaptic: Heh, I was hoping this would fly under the radar :flying_saucer: the warning is that `builder.operations` is not defined in the TypeScript (because it's a private property), but this is the best/only way to _remove_ operations from a `TransactionBuilder`. The correct approach, I think, would be to do something like `{ ignoreOperations: true }` as an option to `cloneFrom`, but that would be a "feature request", so this is an acceptable workaround imo, especially given how much better this is than the existing solution anyway.

sreuland: got it. another option for consideration in future, rather than specific to cloning, i've seen a few other builder patterns with a method to clear the builder state also, like `builder.clearOperations()` which then can setup for using `builder.addOperation()` again.

Shaptic: Added a comment to clarify!

Shaptic: Good point, that may be a cleaner approach! A reset button of sorts.

PR #123, opened 2023-08-10
Fixes `SimulateTransactionResponse` error field

Shaptic:
It's just a string, not an RPC error: https://soroban.stellar.org/api/methods/simulateTransaction.

### DIFF
--- a/src/soroban_rpc.ts
+++ b/src/soroban_rpc.ts
@@ -127,7 +127,7 @@ export namespace SorobanRpc {
 
   export interface SimulateTransactionResponse {
     id: string;
-    error?: jsonrpc.Error;
+    error?: string;
     // this is SorobanTransactionData XDR in base64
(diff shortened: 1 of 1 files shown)


tsachiherman (approved):

src/soroban_rpc.ts
```diff
@@ -127,7 +126,7 @@ export namespace SorobanRpc {
 
   export interface SimulateTransactionResponse {
     id: string;
-    error?: jsonrpc.Error;
+    error?: string;
```
sreuland, line 129: ah, we don't have much in tests that exercise the ts bindings.

sreuland (approved):

src/soroban_rpc.ts
Shaptic: nope, no "integration" tests unfortunately

PR #129, opened 2023-08-16
Automatically decode XDR values in `getTransaction`

Shaptic:
The RPC response schemas for retrieving transaction details have been upgraded to parse the base64-encoded XDR automatically. Succinctly,
  - `GetTransactionResponse` -> `RawGetTransactionResponse`
  - Now, `GetTransactionResponse` includes the full, decoded XDR structures instead of raw, base64-encoded strings for the relevant fields

Here is the full list of changed fields:

```diff
  export interface GetTransactionResponse {
-    envelopeXdr?: string;
-    resultXdr?: string;
-    resultMetaXdr?: string;
+    envelopeXdr?: xdr.TransactionEnvelope;
+    resultXdr?: xdr.TransactionResult;
+    resultMetaXdr?: xdr.TransactionMeta;
+    returnValue?: xdr.ScVal;  // present iff resultMeta is a v3 and has sorobanData
  }
```

Notice that we also provide a property that is a parsed `returnValue` out of the meta, if possible. The true, raw, underlying schema is still accessible by adding the `Raw` prefix to the interface name.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -287,18 +287,46 @@ export class Server {
    *
    * @param {string} hash - The hex-encoded hash of the transaction to check.
    *
-   * @returns {Promise<SorobanRpc.GetTransactionResponse>} Returns a
-   *    promise to the {@link SorobanRpc.GetTransactionResponse} object
-   *    with the status, result, and other details about the transaction.
+   * @returns {Promise<SorobanRpc.GetTransactionResponse>} Returns a promise to
+   *    the {@link SorobanRpc.GetTransactionResponse} object with the status,
+   *    result, and other details about the transaction. Raw XDR fields are
+   *    parsed into their appropriate structures wherever possible.
    */
   public async getTransaction(
     hash: string,
   ): Promise<SorobanRpc.GetTransactionResponse> {
-    return await jsonrpc.post(
+    return await jsonrpc.post<SorobanRpc.RawGetTransactionResponse>(
       this.serverURL.toString(),
       "getTransaction",
       hash,
-    );
+    ).then(raw => {
+      let successInfo: any = {};
+
+      if (raw.status === 'SUCCESS') {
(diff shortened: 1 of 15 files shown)


src/soroban_rpc.ts
```diff
@@ -63,6 +63,24 @@ export namespace SorobanRpc {
     oldestLedger: number;
     oldestLedgerCloseTime: number;
 
+    // the fields below are set if status is SUCCESS
+    ledger?: number;
```
aristidesstaffieri, line 67: Maybe this should be a union type with of `GetTransactionResponseSuccess | GetTransactionResponseError`?

aristidesstaffieri (approved):

src/server.ts
```diff
@@ -287,18 +287,46 @@ export class Server {
...
+   * @returns {Promise<SorobanRpc.GetTransactionResponse>} Returns a promise to
+   *    the {@link SorobanRpc.GetTransactionResponse} object with the status,
+   *    result, and other details about the transaction. Raw XDR fields are
+   *    parsed into their appropriate structures wherever possible.
    */
   public async getTransaction(
     hash: string,
   ): Promise<SorobanRpc.GetTransactionResponse> {
-    return await jsonrpc.post(
+    return await jsonrpc.post<SorobanRpc.RawGetTransactionResponse>(
```
paulbellamy, line 298: Nit: This fn would be much easier to read if you used async/await instead of mixing it with `.then` continuations...

```ts
const raw = await jsonrpc.post<SorobanRpc.RawGetTransactionResponse>(/* ... */);
let successInfo: any = {};
if (raw.status === 'SUCCESS') {
  // ...
}

// ...
```

paulbellamy (approved):

src/server.ts
Shaptic: nice, fixed up!

PR #131, opened 2023-08-21
fix: add defaults when parsing simulation

willemneal:
Fixes issues with a contract call that returns an error type.

### DIFF
--- a/src/transaction.ts
+++ b/src/transaction.ts
@@ -134,13 +134,22 @@ export function parseRawSimulation(
     // Gordon Ramsey in shambles
     return sim;
   }
+  if (!sim.events) {
+    sim.events = []
+  }
+  let transactionData;
+  if (sim.transactionData) {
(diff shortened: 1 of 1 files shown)


src/transaction.ts
```diff
@@ -134,13 +134,22 @@ export function parseRawSimulation(
...
   }
+  if (!sim.events) {
+    sim.events = []
+  }
+  let transactionData;
+  if (sim.transactionData) {
+    transactionData = new SorobanDataBuilder(sim.transactionData)
+  } else {
+    transactionData = new SorobanDataBuilder();
+  }
```
Shaptic, line 145: Isn't this behavior implied?

https://github.com/stellar/js-stellar-base/blob/006f2c97410a39d73576c23951ad8aab8a1aa6b5/src/sorobandata_builder.js#L40

Shaptic (approved):
:clap:

PR #132, opened 2023-08-23
Handle new RPC simulation response variations

Shaptic:
This aligns with https://github.com/stellar/soroban-tools/pull/865 and https://github.com/stellar/soroban-tools/pull/878. The new behavior is documented [here](https://soroban.stellar.org/api/methods/simulateTransaction#returns). In summary:

The `id` and `latestLedger` fields **always** exist. Then, there are branching variations:
   - in error cases:
     * `error` always exists 
   - in success cases:
     * `transactionData`, `minResourceFee`, and `cost` always exist
     * `results` only exists IF the simulation was for an invocation op
     * `restorePreamble` only exists IF the simulation required expired entries
       - this means a restoration is required for the invocation to exist, and this object contains both the fee andfootprint necessary for that
  - `events` **may** exist but we will make it always exist (maybe empty)

Our transformations are as follows:
  * the interface is broken down into types based on success / failure
  * every XDR object is decoded when possible
  * `events` always exists (but may be empty)
  * `results?[]` is coalesced into a single `result?`


src/soroban_rpc.ts
```diff
@@ -162,16 +162,77 @@ export namespace SorobanRpc {
...
+
+    /**
+     * Indicates that a restoration is necessary prior to submission.
+     *
+     * In other words, seeing a restoration preamble means that your invocation
+     * was executed AS IF the required ledger entries were present, and this
+     * field includes information about what you need to restore for the
+     * simulation to succeed.
+     */
+    restore?: {
```
paulbellamy, line 215: Why rename this from `restorePreamble`? The other name seems fine and less confusing when comparing with the docs.

src/soroban_rpc.ts
```diff
@@ -162,16 +162,77 @@ export namespace SorobanRpc {
...
+   *     if an invocation was simulated (since otherwise there's nothing to
+   *     "resultify")
+   *   - if there was an expiration error, this includes error and restoration
+   *     fields
+   *   - for all other errors, this only includes error fields
+   *
+   * @see https://soroban.stellar.org/api/methods/simulateTransaction#returns
+   */
+  export type SimulateTransactionResponse =
+    | SimulateTransactionSuccessResponse
```
sreuland, line 178: great design, how do you feel about sub-classing `SimulateSuccessResponse` a bit further to capture those other conditional data models so clients aren't (re)implementing same on their side?
```
SimulateInvokeHostFnResponse extends SimulateSuccessResponse
SimulateExpiredResponse extends SimulateSuccessResponse
```

`SimulateSuccessResponse` still used as catch-all on remainder of success sim responses that have no further differentiation.

Shaptic: I like the intention, but TypeScript has a **very** annoying limitation in that you can't just do 

```typescript
if (response instanceof SimulateExpiredResponse) { /* ... handle it */ }
```

like you can in other languages, since interfaces are basically only for transpile-time type checking. So even if we added these, users would **still** need to do the equivalent of

```typescript
if (isSimulateExpiredResponse(response)) { /* ... handle it */ }
```

whose implementation (as you can see [here](https://github.com/stellar/js-soroban-client/pull/132/files#diff-fff0ee1c855d7fdef5ec027268d27066a50d0c0dc4e4832b51f8a2e40129e177R192-R198)) is just a bunch of fragile "does it have this property" conditionals.

Plus, clients would have to have way more checks to "break down" the union type (i.e. lots more like [this diff](https://github.com/stellar/js-soroban-client/pull/132/files#diff-fff0ee1c855d7fdef5ec027268d27066a50d0c0dc4e4832b51f8a2e40129e177R55-R57) in the PR). (Then again, you could just `as <ExpectedType>` it, but that's also annoying / fragile.) I think it may be, for now, a little more user-friendly to do

```typescript
if (!!response.restorePreamble) { /* ... handle it */ }
```

as it explicitly marks the field that the client can expect to be present.

My rationale is primarily motivated by the use-case here: I think that users know which fields to expect in which cases. For example, if you **know** you're simulating an invocation, you _also_ know that there will be a `result` field in the success case. Then, you _also_ know that you should check for the `restorePreamble` field prior to submission. So there is less of a need to break down the success union type.

I guess if we had more response variations like this in the future (e.g. more fields or branches), I would be more in support of the interface breakdown. But as such, it might be types for the sake of types at the cost of unnecessary client-side code.

Shaptic: Actually, I added `isSimulationRestore` in a8e2001 to break it down at least a bit further.

Shaptic:
This is targeting the `testnet` branch which will contain the updated XDR changes. I want to land this first before trying to incorporate those, hence the older XDR schema in this PR.

src/soroban_rpc.ts
```diff
@@ -162,16 +162,90 @@ export namespace SorobanRpc {
...
-    transactionData: SorobanDataBuilder;
+
+    /** always present: the LCL known to the server when responding */
+    latestLedger: number;
+
+    /**
+     * The field is always present, but may be empty in cases where:
+     *   - you didn't simulate an invocation,
+     *   - there were no events, or
+     *   - you are communicating with an RPC server w/o diagnostic events on
```
sreuland, line 193: I think can remove ln 193 in terms of events with simulations, as should always get diag events back from simulation, because diag level is hardcoded in the preflight rpc endpoint when it configs host lib - https://github.com/stellar/soroban-tools/blob/main/cmd/soroban-rpc/lib/preflight/src/preflight.rs#L51

it looks like it doesn't go through ingestion with captive core in rpc which is where diag events could be (dis)abled through captive .cfg, I hope I'm stating this accurately.

src/transaction.ts
```diff
@@ -52,13 +52,13 @@ export function assembleTransaction(
     );
   }
 
-  const coalesced = parseRawSimulation(simulation);
-  if (!coalesced.result) {
-    throw new Error(`simulation incorrect: ${JSON.stringify(coalesced)}`);
+  let success = parseRawSimulation(simulation);
```
sreuland, line 55: minor nit, your call, it's not success yet, maybe `simulationResponse`

src/transaction.ts
```diff
@@ -134,40 +134,80 @@ export function parseRawSimulation(
...
+
+  return (typeof sim.error === 'string')
+    // error type: just has error string
+    ? {
+      ...base,
+      error: sim.error,
+    }
+    // success type: might have a result (if invoking) and might have a
+    // restoration hint (if invoking AND some state is expired)
+    : {
```
sreuland, line 155: seems like a lot to inline, just for consideration, a little refactor:

```
{
  if (typeof sim.error === 'string')
      // error type: just has error string
      return {
        ...base,
        error: sim.error,
      }
  }
  
  return assembleSuccess(base, sim);
}

function assembleSuccess(base, sim) {
  // lines 153-187
}
```

test/unit/server/simulate_transaction_test.js
```diff
@@ -136,34 +87,60 @@ describe('Server#simulateTransaction', function () {
...
     // FIXME: This is a workaround for an xdrgen bug that does not allow you to
-    // build "perfectly-equal" xdr.ExtensionPoint instances (but they will still
-    // be binary-equal, so the test passes), but it should be fixed once we
-    // upgrade the XDR to the final testnet version.
+    // build "perfectly-equal" xdr.ExtensionPoint instances (but they're still
+    // binary-equal, so the test passes).
     parsedCopy.transactionData = parsedCopy.transactionData.build();
     parsed.transactionData = parsed.transactionData.build();
 
     expect(parsed).to.be.deep.equal(parsedCopy);
```
sreuland, line 112: these tests that parse raw into type specific sim response, also seem like a good place to add coverage on those new helper functions:

```
expect(IsSimulationSuccess(parsed)).to.equal(true);

```

sreuland (approved):
looks good, left minor suggestions for consideration.

src/transaction.ts
Shaptic: Great point, should have refactored this once it was stable. Done in 301329f8b1ebcd7c75843cb0ccc2dcf2a92282dd.

src/transaction.ts
Shaptic: but it should be! heh. slightly more annoying to read later lines if it's something like `parsed`, and reassigning after the `if (!isSuccess...)` check is also slightly more annoying :upside_down_face:

PR #136, opened 2023-09-06
Add fallback parsing to edge cases in `simulateTransaction`.

Shaptic:
This works around a Soroban RPC bug (https://github.com/stellar/soroban-tools/issues/933) in which `results` will have a result object that is invalid. The `parseRawSimulation` abstraction assumes a valid schema, so this changes `retval` to be a `void` `ScVal` in case of an invalid response.

This also adds a way to make the raw JSON RPC request again, `Server._simulateTransaction()`, in order to prevent any internal parsing errors from affecting development. Developers can then just look at the JSON directly (it also helps with debugging).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,15 @@ A breaking change should be clearly marked in this log.
 ## Unreleased
 
 
+## v0.11.3
+
+### Fixed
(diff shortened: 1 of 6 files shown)


Shaptic:
Duped in #132 now.

PR #138, opened 2023-09-08
fix(spec): unwrap result when parsing return value

willemneal:
This PR unwraps the result type's ok type so that it can be parsed correctly. The error type is used when if it throws.
fixes: https://github.com/stellar/soroban-tools/issues/935

### DIFF
--- a/src/contract_spec.ts
+++ b/src/contract_spec.ts
@@ -146,7 +146,7 @@ export class ContractSpec {
     if (output.switch().value === xdr.ScSpecType.scSpecTypeResult().value) {
       return this.scValToNative(
         val,
-        (output.value() as xdr.ScSpecTypeResult).okType()
+        output.result().okType()
       );
     }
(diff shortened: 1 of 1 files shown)


src/contract_spec.ts
```diff
@@ -143,6 +143,12 @@ export class ContractSpec {
       throw new Error(`Multiple outputs not supported`);
     }
     let output = outputs[0];
+    if (output.switch().value === xdr.ScSpecType.scSpecTypeResult().value) {
+      return this.scValToNative(
+        val,
+        (output.value() as xdr.ScSpecTypeResult).okType()
```
Shaptic, line 149: you can use the appropriate accessor to not need this typecast: 

```suggestion
        output.result().okType()
```

Shaptic (approved):
one small non-blocking nit, otherwise lgtm! a changelog entry would also be appreciated

PR #135, opened 2023-09-01
Upgrade the codebase to work with testnet XDR

Shaptic:

PR #142, opened 2023-09-12
Merge testnet upgrades into `main`

Shaptic:
This is just the meta-PR tracking changes to `testnet`. So far there's just:

- [x] #135 
- [x] #132 pending
- [x] #143 pending


CHANGELOG.md
```diff
@@ -8,6 +8,17 @@ A breaking change should be clearly marked in this log.
 ## Unreleased
 
 
+## v1.0.0-beta.0
+**Note:** This version is currently only compatible with standalone networks built with stellar/stellar-xdr@9ac0264 (the Protocol 20 preview).
```
sreuland, line 12:
```suggestion
**Note:** This version is currently only compatible with stellar networks that are running Protocol 20, such as your own standalone network, futurenet, or testnet.
```

Shaptic: but this isn't true yet :confused: I was going to amend this line once we actually deployed to futurenet/testnet

sreuland (approved):
looks good, I was on the prior pr's so not digging in much, left one comment for consideration on changelog notes.

CHANGELOG.md
sreuland: ah ok, it's not immediately clear what `stellar/stellar-xdr@9ac0264` equates to as it's a low level component that users may not be familiar, maybe `with networks running the Soroban Preview 11 release`. Then wouldn't need to amend it later?

Shaptic: Yeah, I wasn’t sure what the best way to pin the exact XDR/Core version this would work with would be. Preview 11 is nice but it also “doesn’t exist” yet? But maybe that’s a good thing, as people would realize that it’s for a pending release. I’ll amend it!

PR #146, opened 2023-09-14
Properly parse `simulateTransaction` response variations

Shaptic:
This modifies the response parsing to simply attach a `_parsed` field onto the parsed structure, simplifying the error-prone schema detection mechanism from before.

It also updates various TypeScript definitions with the correct field types: latestLedger is a string, not a number, in most responses.

### DIFF
--- a/src/transaction.ts
+++ b/src/transaction.ts
@@ -121,6 +121,7 @@ export function parseRawSimulation(
 
   // shared across all responses
   let base: SorobanRpc.BaseSimulateTransactionResponse = {
+    _parsed: true,
     id: sim.id,
     latestLedger: sim.latestLedger,
     events: sim.events?.map(
@@ -171,7 +172,7 @@ function parseSuccessful(
     )
   };
 
-  if (!sim.restorePreamble) {
+  if (!sim.restorePreamble || sim.restorePreamble.transactionData === '') {
     return success;
   }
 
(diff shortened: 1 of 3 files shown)


src/transaction.ts
```diff
@@ -187,29 +188,12 @@ function parseSuccessful(
...
-      asGud.result !== undefined ||
-      typeof asGud.transactionData !== 'string'
-    ) ||
-    (asRaw.error !== undefined && (
-      !asRaw.events?.length ||
-      typeof asRaw.events![0] === 'string'
-    )) ||
-    (asRaw.results ?? []).length > 0
-  );
+  return !(sim as SorobanRpc.SimulateTransactionResponse)._parsed;
```
sreuland, line 196: nice!

test/unit/server/simulate_transaction_test.js
```diff
@@ -235,3 +237,43 @@ function invokeSimulationResponseWithRestoration(address) {
     }
   };
 }
+
+describe('works with real responses', function() {
```
sreuland, line 241: not suggesting a change, just wondering on hardcoding xdr as precedence in tests or build the js xdr object graph and then serialize via `.toXdr()` rather than hardcoding?

sreuland (approved):

test/unit/server/simulate_transaction_test.js
Shaptic: honestly I mostly just added this as a sanity check that the specific error we were seeing got fixed. another nice part about this is that if the XDR changes, this test will fail, prompting investigation and tests with a new RPC

I agree in general that it's better to build XDR then parse it rather than hardcoding, but that also doesn't reflect real RPC behavior sometimes.. we really need to build integration tests for RPC :arrow_right: soroban-client, but now isn't the time for that heh

sreuland: agreed, this isn't the place for that type of integration test, this should just work against the rpc interface, that's what system-test/e2e is for.

Issue #128, opened 2023-08-14
Decode XDR structures wherever possible.


Shaptic:
**Is your feature request related to a problem? Please describe.**
It's annoying to decode base64-encoded XDR that comes from the raw JSON responses because:

 (a) you don't necessarily know what the type is without referencing the documentation
 (b) you need to actually write the decoding code and error-check it

**Describe the solution you'd like**
This should be part of `Server`s request methods: they should wrap and decode the raw responses wherever possible:

- [x] simulateTransaction
- [x] getTransaction
- [ ] sendTransaction
- [x] getLedgerEntries
- [x] getLatestLedger
- [x] getEvents

**Additional context**
The work for this has been started in #127 and #129. 

**Please :+1: or :-1: as feedback for continuing this feature!**

PR #154, opened 2023-10-03
Automatically decode XDR for `getLedgerEntries` and friends

Shaptic:
### What
We have the following breaking schema change in the top-level response: 

```diff
-entries?: LedgerEntryResult[];
+entries: LedgerEntryResult[];
```

and in each entry:

```diff
-key: string;
-xdr: string;
+key: xdr.LedgerKey;
+val: xdr.LedgerEntryData;
```

More specificially,
 - `entries` is now guaranteed to exist, but it may be empty
 - `entries[i].key` is an instance of `xdr.LedgerKey`
 - the `entries[i].xdr` field is now `val`, instead
 - `entries[i].val` is an instance of `xdr.LedgerEntryData`

If users want to continue to use the raw RPC response, there is still a `Server._getLedgerEntries` method.

Some helpers from previous PRs have been moved around into a new `src/parsers.ts`, also.

### Why
Better UX, see #128.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -8,7 +8,6 @@ export { ContractSpec } from "./contract_spec";
 export { Server, Durability } from "./server";
 export { AxiosClient, version } from "./axios";
 export * from "./transaction";
-export * from "./parsers";
 
 // expose classes and functions from stellar-base
 export * from "stellar-base";
--- a/src/index.ts
+++ b/src/index.ts
@@ -9,6 +9,9 @@ export { Server, Durability } from "./server";
 export { AxiosClient, version } from "./axios";
 export * from "./transaction";
 
+// only needed for testing
+export { parseRawSimulation } from "./parsers";
+
 // expose classes and functions from stellar-base
 export * from "stellar-base";
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -9,7 +9,7 @@ export { Server, Durability } from "./server";
 export { AxiosClient, version } from "./axios";
 export * from "./transaction";
 
-// only needed for testing
+// export is necessary for testing, but it may be useful for others
 export { parseRawSimulation } from "./parsers";
 
 // expose classes and functions from stellar-base
--- a/src/index.ts
+++ b/src/index.ts
@@ -5,9 +5,10 @@ export * from "./soroban_rpc";
 export { ContractSpec } from "./contract_spec";
 
 // stellar-sdk classes to expose
-export { Server } from "./server";
+export { Server, Durability } from "./server";
 export { AxiosClient, version } from "./axios";
 export * from "./transaction";
+export * from "./parsers";
 
 // expose classes and functions from stellar-base
 export * from "stellar-base";
--- a/src/parsers.ts
+++ b/src/parsers.ts
@@ -1,7 +1,24 @@
 import { xdr, SorobanDataBuilder } from 'stellar-base';
 import { SorobanRpc } from './soroban_rpc';
 
+export function parseLedgerEntries(
+  raw: SorobanRpc.RawGetLedgerEntriesResponse
+): SorobanRpc.GetLedgerEntriesResponse {
+  return {
+    latestLedger: raw.latestLedger,
+    entries: (raw.entries ?? []).map(rawEntry => {
+      if (!rawEntry.key || !rawEntry.xdr) {
+        throw new TypeError(`invalid ledger entry: ${rawEntry}`);
+      }
 
+      return {
+        lastModifiedLedgerSeq: rawEntry.lastModifiedLedgerSeq,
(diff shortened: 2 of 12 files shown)


src/parsers.ts
```diff
@@ -0,0 +1,110 @@
+import { xdr, SorobanDataBuilder } from 'stellar-base';
+import { SorobanRpc } from './soroban_rpc';
+
+export function parseLedgerEntries(
```
sreuland, line 4: minor for consideration, on the parsing methods, try a consistent naming pattern like `parseRPC<ResponseType>Response` rather than 'raw' or no prefix, etc.

sreuland, line 4: since this method is exported, shoiuld it get jsdoc'd?

sreuland: ah, I see 'Raw' is used consistently further down, maybe just this method follows same pattern like  `parseRawLedgerEntries`

src/soroban_rpc.ts
```diff
@@ -25,6 +25,12 @@ export namespace SorobanRpc {
   }
 
   export interface LedgerEntryResult {
+    lastModifiedLedgerSeq?: number;
+    key: xdr.LedgerKey;
+    val: xdr.LedgerEntryData;
+  }
+
+  export interface RawLedgerEntryResult {
```
sreuland, line 33: minor, could add js-docs with link to public copy rather than inline comments:
```
/*
* See {@link https://soroban.stellar.org/api/methods/getLedgerEntries#return RPCLedgerEntryResult} 
*/
```

src/parsers.ts
Shaptic: ugh, I only export it for testing purposes, so it's not valuable to document (plus I'm too lazy)

src/soroban_rpc.ts
Shaptic: nice, done in e7ed703

src/parsers.ts
Shaptic: good call, fixed up! e7ed703

Shaptic: oh wait, it's not exported on a library level, only in the module file, so no need to jsdoc anyway

src/index.ts
```diff
@@ -5,10 +5,13 @@ export * from "./soroban_rpc";
 export { ContractSpec } from "./contract_spec";
 
 // stellar-sdk classes to expose
-export { Server } from "./server";
+export { Server, Durability } from "./server";
 export { AxiosClient, version } from "./axios";
 export * from "./transaction";
 
+// only needed for testing
```
sreuland, line 12: once it's available in the wild, nothing prevents clients from using in their production side code, should comment just be removed or not exported?

test/unit/server/get_account_test.js
```diff
@@ -71,20 +64,13 @@ describe('Server#getAccount', function () {
...
-          ]
-        ]
+        params: [[key.toXDR('base64')]]
       })
       .returns(
         Promise.resolve({
           data: {
             result: {
+              latestLedger: 0,
               entries: null
```
sreuland, line 74: shouldn't this now be guaranteed at least empty, `entries: []`

tsconfig.json
```diff
@@ -9,7 +9,7 @@
     "moduleResolution": "node",
     "rootDir": "src",
     "outDir": "lib",
-    "target": "es5"
+    "target": "es6"
```
sreuland, line 12: just for consider, could change `lib: ["ES6"]` [at same time](https://www.typescriptlang.org/tsconfig#lib) just to visually drive the two are aligned.

sreuland (approved):
nice work!

test/unit/server/get_account_test.js
Shaptic: yes but the RPC endpoint can still return null

tsconfig.json
Shaptic: good catch and good call! :clap:

test/unit/server/get_account_test.js
sreuland: ok, cool, I was basing off the statement in pr description, may have been different context.

`entries is now guaranteed to exist, but it may be empty`

src/index.ts
Shaptic: I can't avoid exporting because it would mean it isn't available for browser testing, unfortunately :( you're right that nothing is stopping people from using it if they want. I added a comment to that effect in 47e445e.

test/unit/server/get_account_test.js
Shaptic: yeah so that applies to the `Server.getLedgerEntries` _method_, while this is simulating the raw RPC response (which **will** be transformed into having `entries: []`)

PR #156, opened 2023-10-06
Automatically decode XDR structures for `Server.getEvents`

Shaptic:
This makes the following changes to the expected schema for [getEvents endpoint](https://soroban.stellar.org/api/methods/getEvents):

```diff
-   contractId: string;
+   contractId: Contract;
-   topic: string[];
+   topic: xdr.ScVal[];
-   value: {
-     xdr: string;
-   };
+   value: xdr.ScVal;
```

In particular,
 - `events[i].contractId` is now an instance of [`Contract`](https://stellar.github.io/js-soroban-client/Contract.html)
 - `events[i].topic` is now a list of decoded `xdr.ScVal` instances
 - `events[i].value.xdr` is now remapped directly to `events.value`
 - `events[i].value` is a decoded `xdr.ScVal` instance

Related: #128.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -302,47 +302,53 @@ export class Server {
   public async getTransaction(
     hash: string,
   ): Promise<SorobanRpc.GetTransactionResponse> {
-    const raw = await jsonrpc.post<SorobanRpc.RawGetTransactionResponse>(
+    return this._getTransaction(hash).then(raw => {
+      let successInfo: Omit<
+        SorobanRpc.GetSuccessfulTransactionResponse,
+        keyof SorobanRpc.GetFailedTransactionResponse
+      > = {} as any;
+
+      if (raw.status === SorobanRpc.GetTransactionStatus.SUCCESS) {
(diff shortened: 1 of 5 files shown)


src/server.ts
```diff
@@ -388,14 +392,18 @@ export class Server {
...
     // The difficulty comes in matching up the correct integer primitives.
-    return jsonrpc.postObject(this.serverURL.toString(), "getEvents", {
-      filters: request.filters ?? [],
-      pagination: {
-        ...(request.cursor && { cursor: request.cursor }), // add if defined
-        ...(request.limit && { limit: request.limit }),
-      },
-      ...(request.startLedger && { startLedger: request.startLedger.toString() }),
-    });
+    return jsonrpc.postObject<SorobanRpc.RawGetEventsResponse>(
```
sreuland, line 395: is there a reason why `_getEventsRaw()` is not used here, following same 'pair of methods' pattern for raw and encoded responses?

sreuland (approved):

src/server.ts
Shaptic: nope, just an oversight. thanks for the catch!

PR #157, opened 2023-10-06
Automatically parse the error result field in `sendTransaction`

### LINKED ISSUE #128: Decode XDR structures wherever possible.

Shaptic:
In particular,

```diff
-    errorResultXdr?: string;
+    errorResult?: xdr.TransactionResult;
```

The error result field, if present, will automatically be decoded.
 
Closes #128.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -16,7 +16,11 @@ import { Friendbot } from "./friendbot";
 import * as jsonrpc from "./jsonrpc";
 import { SorobanRpc } from "./soroban_rpc";
 import { assembleTransaction } from "./transaction";
-import { parseRawSimulation, parseRawLedgerEntries } from "./parsers";
+import {
+  parseRawSendTransaction,
+  parseRawSimulation,
+  parseRawLedgerEntries
+} from "./parsers";
 
 export const SUBMIT_TRANSACTION_TIMEOUT = 60 * 1000;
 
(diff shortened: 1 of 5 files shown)


src/server.ts
```diff
@@ -626,7 +630,13 @@ export class Server {
   public async sendTransaction(
     transaction: Transaction | FeeBumpTransaction,
   ): Promise<SorobanRpc.SendTransactionResponse> {
-    return await jsonrpc.post(
+    return this._sendTransaction(transaction).then(parseRawSendTransaction);
```
sreuland, line 633: for consideration, a test to demonstrate this new expectation, like `it provides an encoded error result`  in https://github.com/stellar/js-soroban-client/blob/decode-send-transaction/test/unit/server/send_transaction_test.js#L38

sreuland (approved):
looks good, left one comment to consider including some test coverage.

src/server.ts
Shaptic: good call :+1:

Shaptic: done in 930ef87!

PR #158, opened 2023-10-09
Upgrade `stellar-base` and adapt to its breaking changes

Shaptic:


test/unit/server/simulate_transaction_test.js
```diff
@@ -122,8 +122,8 @@ describe('Server#simulateTransaction', function () {
     expect(SorobanClient.SorobanRpc.isSimulationSuccess(parsed)).to.be.true;
   });
 
-  xit('works with restoration', function () {
-    const simResponse = invokeSimulationResponseWithRestoration(address);
+  xit('works with restoration', async function (done) {
```
sreuland, line 125: just checking, should this test still be skipped or enabled, if still disabled, maybe a comment on what's blocking it?

sreuland (approved):

test/unit/server/simulate_transaction_test.js
Shaptic: ooh, great catch, tysm :pray:

it was failing because the fake response generator was non-deterministic. fixed in 06dfff9

PR #160, opened 2023-10-09
Adds a raw `Server._simulateTransaction` to match others

Shaptic:
These are useful for debugging in situations where parsing reveals some underlying bug.

It also removes the `json.post<T>` syntax as the `T` is implied by the `Server` method's return value.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -269,21 +269,19 @@ export class Server {
   public async getLedgerEntries(
     ...keys: xdr.LedgerKey[]
   ): Promise<SorobanRpc.GetLedgerEntriesResponse> {
-    return this._getLedgerEntries(
-      ...expandRequestIncludeExpirationLedgers(keys)
-    )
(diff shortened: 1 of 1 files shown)


src/server.ts
```diff
@@ -272,7 +272,7 @@ export class Server {
   public async _getLedgerEntries(
     ...keys: xdr.LedgerKey[]
   ): Promise<SorobanRpc.RawGetLedgerEntriesResponse> {
-    return jsonrpc.post<SorobanRpc.RawGetLedgerEntriesResponse>(
```
sreuland, line 275: that's odd, i'm confused by the diff display as this wasn't changed - https://github.com/stellar/js-soroban-client/blob/raw-methods/src/server.ts#L274

sreuland (approved):

src/server.ts
Shaptic: odd! maybe it's because the branch was slightly out of date

PR #161, opened 2023-10-10
Include expiration entry on ledger entries in getContractData/getLedgerEntries responses

### LINKED ISSUE #153: Add ability to get expiration information
**Is your feature request related to a problem? Please describe.**
There's no obvious way to get the expiration ledger anymore.

**Describe the solution you'd like**
`Server.getLedgerEntries` et al. should allow a `fetchExpiration` parameter that will automatically include a fetch for the corresponding expiration ledger entry.

**Describe alternatives you've considered**
You can do this manually, but you need to know oddly-specific internals about Stellar to achieve it.

This could be done on the RPC side, but it may take longer and this can be a stop-gap.

**Additional context**
[Workaround explanation](https://github.com/stellar/soroban-react-atomic-swap/pull/13#discussion_r1333384134).
[Discord context](https://discord.com/channels/897514728459468821/1158426302777405600/1158496027612041227).

sreuland:
This is a client-side approach to eagerly fetching the ledger entry expiration info(ExpirationEntry) automatically for any `LedgerDataEntry` requested through getLedgerEntries and getContractData, and then it embeds the ExpirationEntry found as `expiration` on each response item built from the LedgerDataEntry.

it's meant as interim solution until https://github.com/stellar/soroban-tools/issues/1010 modifes the rpc server to include expiration entry info by default on ledger data entries in rpc responses.

Closes #153


src/server.ts
```diff
@@ -749,3 +753,77 @@ function findCreatedAccountSequenceInTransactionMeta(
...
+    if (key.switch().value !== xdr.LedgerEntryType.expiration().value) {
+      const expirationKey = xdr.LedgerKey.expiration(
+        new xdr.LedgerKeyExpiration(
+          { keyHash: hash(key.toXDR())}
+      ));
+      includingExpiryKeys.push(expirationKey);
+    }
+    includingExpiryKeys.push(key);
+  });
+  return includingExpiryKeys;
```
Shaptic, line 828: Since we are always returning the entire `keys` array, it might be a little more idiomatic to do concat + filter + map like so:

```typescript
return keys.concat(keys
  .filter(key => { /* exclude expiration keys */ })
  .map(key => { /* do hashing stuff */ });
```

src/server.ts
```diff
@@ -749,3 +753,77 @@ function findCreatedAccountSequenceInTransactionMeta(
...
+
+
+// TODO - remove once rpc updated to
+// append expiration entry per data LK requested onto server-side response 
+// https://github.com/stellar/soroban-tools/issues/1010
+async function mergeResponseExpirationLedgers(response: Promise<SorobanRpc.RawGetLedgerEntriesResponse>,
+  requestedKeys: xdr.LedgerKey[]
+): Promise<SorobanRpc.RawGetLedgerEntriesResponse> {
+  let ledgerEntriesResponse = await response;
+  const expirationKeyToRawEntryResult:Map<String, SorobanRpc.RawLedgerEntryResult> = new Map();
```
Shaptic, line 765: I think it'd be better to just use the response directly rather than the promise and put the `await` in the higher level (i.e. in `Server.getLedgerEntries`) - then this can be used & tested synchronously! 

```suggestion
function mergeResponseExpirationLedgers(
  response: SorobanRpc.RawGetLedgerEntriesResponse,
  requestedKeys: xdr.LedgerKey[]
): SorobanRpc.RawGetLedgerEntriesResponse {
  const ledgerEntriesResponse = response;
  const expirationKeyToRawEntryResult:Map<String, SorobanRpc.RawLedgerEntryResult> = new Map();
```

src/server.ts
```diff
@@ -749,3 +753,77 @@ function findCreatedAccountSequenceInTransactionMeta(
...
+  const requestedKeyXdrs = new Set<String>(requestedKeys.map(requestedKey =>
+    requestedKey.toXDR('base64')));
+
+  (ledgerEntriesResponse.entries ?? []).forEach(rawEntryResult => {
+    if (!rawEntryResult.key) {
+      // don't interpret here, just pass it through to the outgoing response as-is
+      expirationKeyToRawEntryResult.set(Math.random().toString(), rawEntryResult)
+      return;
+    }
+    const parsedKey:xdr.LedgerKey = xdr.LedgerKey.fromXDR(rawEntryResult.key, 'base64');
```
Shaptic, line 775: nit: type is _heavily_ implied here, unneeded verbosity
```suggestion
    const parsedKey = xdr.LedgerKey.fromXDR(rawEntryResult.key, 'base64');
```

test/unit/server/get_ledger_entries_test.js
```diff
@@ -0,0 +1,262 @@
...
+        let result = response.entries[0];
+        expect(result.key.toXDR('base64')).to.be.deep.equal(
+          expectedResult.key.toXDR('base64')
+        );
+        expect(result.val.toXDR('base64')).to.be.deep.equal(
+          expectedResult.val.toXDR('base64')
+        );
+        expect(result.expiration.toXDR('base64')).to.be.deep.equal(
+          expectedResult.expiration.toXDR('base64')
+        );
```
Shaptic, line 98: you don't need to `.deep.equal` on base64 strings as comparing strings is already "deep", or you can compare the objects directly with `.deep` without encoding them into base64!

test/unit/server/get_ledger_entries_test.js
```diff
@@ -0,0 +1,262 @@
...
+        expect(result.val.toXDR('base64')).to.be.deep.equal(
+          expectedResult.val.toXDR('base64')
+        );
+        expect(result.expiration).to.be.undefined;
+        done();
+      })
+      .catch((err) => done(err));
+  });
+});
```
Shaptic, line 262: it looks like the assertions are identical in each of these cases. there's a lot of verbosity in the setup; could we combine the setup code into a single blob and do a loop over the expected/actual scenarios?

great coverage, though!

Shaptic (commented):
quick first-pass review, looks pretty good but hoping we can simplify this? great test coverage to kick it off though :clap:

src/soroban_rpc.ts
```diff
@@ -36,6 +37,8 @@ export namespace SorobanRpc {
     key: string;
     /** a base-64 encoded {@link xdr.LedgerEntryData} instance */
     xdr: string;
+    /** an optional base-64 endoded  {@link xdr.ExpirationEntry} instance*/
+    expiration?: string;
```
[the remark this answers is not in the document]
sreuland: that way it passes through parseRawLedgerEntries seamless to include it on `GetLedgerEntriesResponse`, represents the model change we'd mostly anticipate from rpc to follow, 'contract first' approach?

src/parsers.ts
```diff
@@ -38,13 +38,14 @@ export function parseRawLedgerEntries(
...
       if (!rawEntry.key || !rawEntry.xdr) {
-        throw new TypeError(`invalid ledger entry: ${rawEntry}`);
+        throw new TypeError(`invalid ledger entry: ${JSON.stringify(rawEntry)}`);
       }
 
       return {
         lastModifiedLedgerSeq: rawEntry.lastModifiedLedgerSeq,
         key: xdr.LedgerKey.fromXDR(rawEntry.key, 'base64'),
         val: xdr.LedgerEntryData.fromXDR(rawEntry.xdr, 'base64'),
+        expiration: rawEntry.expiration ? xdr.ExpirationEntry.fromXDR(rawEntry.expiration!, 'base64') : undefined
```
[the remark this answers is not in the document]
sreuland: ha, yeah, I had tried spread here for a bit, but that was the nuance, nice!

package.json
```diff
@@ -35,13 +35,13 @@
     "clean": "rm -rf lib/ dist/ coverage/ .nyc_output/ jsdoc/",
     "docs": "yarn build:docs && jsdoc -c .jsdoc.json --verbose",
     "test": "yarn test:node && yarn test:integration && yarn test:browser",
-    "test:node": "yarn build:test && yarn _nyc mocha --recursive 'test/unit/**/*.js'",
-    "test:integration": "yarn _nyc mocha --recursive 'test/integration/**/*.js'",
+    "test:node": "yarn _build && yarn build:test && yarn _nyc mocha --recursive 'test/unit/**/*.js'",
```
sreuland, line 38: noticed locally that running `yarn test:node` often wasn't running latest .ts changes, need to rebuild .ts first

src/server.ts
```diff
@@ -749,3 +753,77 @@ function findCreatedAccountSequenceInTransactionMeta(
 
   throw new Error("No account created in transaction");
 }
+
+
+// TODO - remove once rpc updated to
+// append expiration entry per data LK requested onto server-side response 
+// https://github.com/stellar/soroban-tools/issues/1010
+async function mergeResponseExpirationLedgers(response: Promise<SorobanRpc.RawGetLedgerEntriesResponse>,
```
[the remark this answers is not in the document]
sreuland: i looked into the rpc handler, https://github.com/stellar/soroban-tools/blob/main/cmd/soroban-rpc/internal/db/ledgerentry.go#L252, it ultimately iterates over a map to assemble the final Entries[], and the map iterator is not deterministic to original insert order on the map. so, I took the extra defensive approach

src/server.ts
sreuland: https://github.com/stellar/js-soroban-client/pull/161/commits/e6e944a48b4a8bee1d7edc30e16ffb382d29be56

test/unit/server/get_ledger_entries_test.js
sreuland: thanks, overlooked on that from c&p, fixed.

sreuland:
> quick first-pass review, looks pretty good but hoping we can simplify this? great test coverage to kick it off though 👏

@Shaptic , I've responded to feedback, all commits in, lmk, thanks!

package.json
Shaptic: yeah I did this separation on purpose so that you can iterate on tests faster (you don't need to `yarn build` if you're just changing tests)

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ A breaking change should be clearly marked in this log.
 
 
 ## Unreleased
+* Include `expiration` attribute on ledger data entries in `getContractData` and `getLedgerEntries` responses ([#153](https://github.com/stellar/js-soroban-client/pull/153)).
```
Shaptic, line 9: Thanks for adding this! Can you put it in an `### Added` section?

src/server.ts
```diff
@@ -749,3 +753,77 @@ function findCreatedAccountSequenceInTransactionMeta(
...
+  requestedKeys: xdr.LedgerKey[]
+): Promise<SorobanRpc.RawGetLedgerEntriesResponse> {
+  let ledgerEntriesResponse = await response;
+  const expirationKeyToRawEntryResult:Map<String, SorobanRpc.RawLedgerEntryResult> = new Map();
+  const requestedKeyXdrs = new Set<String>(requestedKeys.map(requestedKey =>
+    requestedKey.toXDR('base64')));
+
+  (ledgerEntriesResponse.entries ?? []).forEach(rawEntryResult => {
+    if (!rawEntryResult.key) {
```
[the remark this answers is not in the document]
Shaptic: ah I see - I think I added that because it happened in testing cases. it should never happen with a real RPC so throwing here is better than a random cache buster, I think? wdyt?

src/server.ts
```diff
@@ -749,3 +752,69 @@ function findCreatedAccountSequenceInTransactionMeta(
...
+      let rawEntryResultLookup = expirationKeyToRawEntryResult.get(keyHash);
+
+      if (!rawEntryResultLookup) {
+        rawEntryResultLookup = rawEntryResult;
+        expirationKeyToRawEntryResult.set(keyHash, rawEntryResultLookup);
+      } 
+      rawEntryResultLookup.key = rawEntryResult.key;
+      rawEntryResultLookup.xdr = rawEntryResult.xdr;
+      rawEntryResultLookup.lastModifiedLedgerSeq = rawEntryResult.lastModifiedLedgerSeq;
+      return;
```
Shaptic, line 787: Sorry, I'm having trouble following this logic. It seems like 

 * if the entry doesn't exist, you're setting `keyHash` to `rawEntryResultLookup` 
   - (which is the same as `rawEntryResult`))
 * then, if it _does_ exist, you are setting its fields to the same as those in `rawEntryResult` (i.e. overwriting what was there anyway)

Doesn't that mean that this whole sequence is always equivalent to just:

```suggestion
      expirationKeyToRawEntryResult.set(rawEntryResult);
      return;
```

Am I misunderstanding something here?

src/server.ts
```diff
@@ -749,3 +752,69 @@ function findCreatedAccountSequenceInTransactionMeta(
...
+  ledgerEntriesResponse.entries = [...expirationKeyToRawEntryResult.values()];
+  return ledgerEntriesResponse;
+}
+
+// TODO - remove once rpc updated to
+// include expiration entry on responses for any data LK's requested  
+// https://github.com/stellar/soroban-tools/issues/1010
+function expandRequestIncludeExpirationLedgers(
+  keys: Array<xdr.LedgerKey>
+): Array<xdr.LedgerKey> {
```
Shaptic, line 815: you can use `[]` here instead of `Array` - that's what the codebase does elsewhere afaict

```suggestion
function expandRequestIncludeExpirationLedgers(keys: xdr.LedgerKey[]): xdr.LedgerKey[] {
```

src/soroban_rpc.ts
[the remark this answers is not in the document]
Shaptic: I see what you're saying but this busts the `_getLedgerEntries` schema that returns the RPC directly, right? Including this in the non-raw / parsed version makes more sense to me. And we can modify this once the RPC is updated.

src/soroban_rpc.ts
```diff
@@ -28,6 +28,7 @@ export namespace SorobanRpc {
     lastModifiedLedgerSeq?: number;
     key: xdr.LedgerKey;
     val: xdr.LedgerEntryData;
+    expiration?: xdr.ExpirationEntry;
```
Shaptic, line 31: should we just return the sequence number? or is there more in the entry that might be useful to users?

test/unit/server/get_ledger_entries_test.js
Shaptic: fair enough!

Shaptic (commented):
getting close, some minor comments and a few Qs about the parsing logic

package.json
sreuland: yep, figured there was nuances, ok, not sure on which is best dx, if you change any app ts code you have to yarn build manually first before running any test targets or you'll get stale/incorrect results, I'll revert unless you feel it's worth changing.

src/server.ts
sreuland: if the order of GetLedgerEntriesResponse.entries was:
`[RawLedgerEntryResult.key=expirationKey, RawLedgerEntryResult.key=datakey]`

then the lookup gets populated with keyHash={expiration: xyz} on first iteration with the expiration entry on line 798, then the data entry comes through on second iteration, produces the same keyHash, gets that existing lookup value and on lines 784-786 its props are merged to form final object of {expiration, key, xdr, lastModifiedLedgerSeq}

i looked around a bit on merging or union of two objects with result of superset of keys, but didn't find much and settled on with this manual thing.

src/soroban_rpc.ts
sreuland: yeah, good point, I wondered about the expiration model aspect, how much this should be a one-one to network or abstract further, I think it is more ideal to abstract up further to app purpose, `expirationLedgerSeq` is the only relevant/actionable fact from an expiration right, no-one needs the keyHash, I'll change this to `expirationLedgerSeq`.

src/soroban_rpc.ts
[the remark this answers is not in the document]
sreuland: I don't understand how it would break compat, `expiration` is optional, current rpc responses which don't have `expiration` shouldn't trigger runtime error when marshaled into this, and `mergeResponseExpirationLedgers` takes in this raw response model from rpc first, populates `expiration` on the model if possible, and forwards to parse method, ideally https://github.com/stellar/soroban-tools/issues/1010 will just update the response model to same, then it's just small change here in client to remove `mergeResponseExpirationLedgers`/`expandRequestIncludeExpirationLedgers`, the rest of this should remain and work the same.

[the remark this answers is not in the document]
sreuland: ah, well, have to change the tests also to no longer expect axiosMock to include expiration keys in requests and expiration entry in responses.

src/soroban_rpc.ts
sreuland: @tsachiherman , as part of https://github.com/stellar/soroban-tools/issues/1010, wanted to confirm on expected changes to [LedgerEntryResult](https://github.com/stellar/soroban-tools/blob/main/cmd/soroban-rpc/internal/methods/get_ledger_entries.go#L21-L28), so that we can pre-plumb the same model into js client here, it will likely have `expirationLedgerSeq` and `isExpired` added correct?

src/soroban_rpc.ts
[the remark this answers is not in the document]
sreuland: I'm going to replace `expiration` with `expirationLedgerSeq` on `RawLedgetEntryResult.expiration` and `LegderEntryResult` per [the review comment](https://github.com/stellar/js-soroban-client/pull/161/files/6cda5badbdef651d8a0da50f51e0b929c3634ffa#r1353836257) which also lines up with what was mentioned for specs on [rpc side/#1010](https://github.com/stellar/soroban-tools/issues/1010), will have the same optional flow here in client for the interim, but ideally client will have small, low-impact cut-over path once present on rpc responses.

src/server.ts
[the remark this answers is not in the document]
sreuland: yes, added test coverage on that also - https://github.com/stellar/js-soroban-client/pull/161/commits/f582dccdb466b52331b947ca32034cb01909be77

src/soroban_rpc.ts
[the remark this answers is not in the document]
Shaptic: okay I see your rationale! I wasn't thinking about marshalling, more-so that the `Raw*` schema should line up exactly with the RPC schema. But since we're getting #1010 out soon, and since the field is optional, I'm fine with including here in advance.

package.json
Shaptic: yeah it depends on your dev flow, too - it's just nice to run tests quickly w/o needing to build. maybe there should be a warning about it being stale? I dunno, but agreed that it's not ideal. reverting sounds good since we both know the quirks now

src/server.ts
Shaptic: ahhhh yes okay. so the spread operator (`...{}`) should help us do a merge, then: it will copy existing keys and you can add new ones

src/soroban_rpc.ts
Shaptic: yeah and this aligns w/ RPC, exactly like you said

src/server.ts
```diff
@@ -749,3 +752,68 @@ function findCreatedAccountSequenceInTransactionMeta(
...
+    requestedKey.toXDR('base64')));
+
+  (ledgerEntriesResponse.entries ?? []).forEach(rawEntryResult => {
+    if (!rawEntryResult.key || !rawEntryResult.xdr) {
+      throw new TypeError(`invalid ledger entry: ${JSON.stringify(rawEntryResult)}`);
+    }
+    
+    const parsedKey = xdr.LedgerKey.fromXDR(rawEntryResult.key, 'base64');
+    
+    if (parsedKey.switch().value !== xdr.LedgerEntryType.expiration().value ||
```
Shaptic, line 774: right okay so I looked at this carefully again and what you want to do is:

- if it's a real ledger entry (not expiration) and key exists, merge with existing entry
- if key not exists, add new entry (w/ key, xdr)
- if an expiration entry, merge or add also

But all of this can be ultra-simplified into just "merge existing (if any) with key hash no matter what":

```typescript
const isExpiration = parsedKey.switch().value === xdr.LedgerEntryType.expiration().value;
const keyHash = (isExpiration) 
  ? parsedKey.expiration().keyHash().toString()
  : hash(parsedKey.toXDR()).toString();

const existing = expirationKeyToRawEntryResult.get(keyHash) ?? {};

if (isExpiration) {
  const expiration = xdr.LedgerEntryData
      .fromXDR(rawEntryResult.xdr, 'base64')
      .expiration()
      .toXDR('base64');
  expirationKeyToRawEntryResult.set(keyHash, { ...existing, expiration });
} else {
  expirationKeyToRawEntryResult.set(keyHash, { ...existing, ...rawEntryResult });
}
```

CHANGELOG.md
sreuland: yes, moved - https://github.com/stellar/js-soroban-client/pull/161/commits/b1a5641ac7a94b72c37bc55cf97e50a867126e5f

src/server.ts
sreuland: yes, great optimization, squeaked it in here https://github.com/stellar/js-soroban-client/pull/161/commits/3b66c37bf87a02c065b4e0aa6e909feed45b46fd

src/soroban_rpc.ts
sreuland: yes, [discussion on #1010](https://github.com/stellar/soroban-tools/issues/1010#issuecomment-1758158987), summary was just a new `expirationLedgerSeq` will be introduced on results, so, updated to align to that model for results here in pr.

src/server.ts
sreuland: yes, thanks for the [optimal suggeestion you made ](https://github.com/stellar/js-soroban-client/pull/161#discussion_r1355618090), which leveraged that and I re-wired to that.

package.json
sreuland: reverted - https://github.com/stellar/js-soroban-client/pull/161/commits/1f483db99ce2c45d566b97d10858cb32418f800f

Shaptic (approved):
awesome work and great test coverage :clap:

PR #164, opened 2023-10-12
update the docs to reference expirationLedgerSeq 

sreuland:
noticed the docs were referring to an older 'expiration' field which was incorrect, it should be `expirationLedgerSeq`


Shaptic (approved):
nice catch!

PR #167, opened 2023-10-24
Prepare release w/ support for stable Protocol 20 and RPC.

Shaptic:
This is a staging ground for all of the changes necessary to get an SDK compatible with the upcoming stable release of Protocol 20. This PR is a conglomeration of PRs that have already been approved:

- [x] #168 
- [x] RPC schema changes: #173
- [x] Port the getEvents fix from #172

Note that this work will also need to be ported to the stellar-sdk.


sreuland (approved):
lgtm, since changing system-test to use js-stellar-sdk, unfortunately won't have e2e coverage on this lib anymore, but is deprecated.

PR #168, opened 2023-10-26
Drop expiration workaround in lieu of new TTL in schema

Shaptic:
The response schema now includes `liveUntilLedgerSeq` by default, so we no longer need to inject the expiration ledgers into requests and parse that information.

(Note that this is targeting the `stable` branch.)

Closes this part of https://github.com/stellar/js-stellar-sdk/issues/869.


stellarsaur (approved):
Nice work, LGTM

PR #171, opened 2023-11-13
feat(ContractSpec): add JsonSchema method for corresponding JSON schema

willemneal:
This will allow tooling like react forms to use the JSON schema to generate UI from the XDR that is valid for the ContractSpec's encoding of args.

### DIFF
--- a/src/contract_spec.ts
+++ b/src/contract_spec.ts
@@ -1,19 +1,9 @@
-import { ScIntType, XdrLargeInt, xdr } from 'stellar-base';
-
-import { Address } from 'stellar-base';
-import { Contract } from 'stellar-base';
-import { scValToBigInt } from 'stellar-base';
-
-export interface Union<T> {
-  tag: string;
-  values?: T;
-}
-
 /**
  * Provides a ContractSpec class which can contains the XDR types defined by the contract.
  * This allows the class to be used to convert between native and raw `xdr.ScVal`s.
  *
  * @example
+ * ```js
  * const specEntries = [...]; // XDR spec entries of a smart contract
  * const contractSpec = new ContractSpec(specEntries);
  *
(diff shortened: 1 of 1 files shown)


src/contract_spec.ts
```diff
@@ -687,11 +765,322 @@ function findCase(name: string) {
...
-function readObj(args: object, input: xdr.ScSpecFunctionInputV0): any {
-  let inputName = input.name().toString();
-  let entry = Object.entries(args).find(([name, _]) => name === inputName);
-  if (!entry) {
-    throw new Error(`Missing field ${inputName}`);
+const PRIMITIVE_DEFINITONS = {
+  U64: {
+    type: 'integer',
+    minimum: 0,
+    maximum: 18446744073709551615
```
leighmcculloch, line 772: This value is precision truncated, it evaluates to this in JS because numbers in JS can only store all integers up to 53-bits, at which point precision is lost:
```
18446744073709552000
```

willemneal: Yeah I was wondering if we should use strings to ensure the precision.  It would be great if JSON had support for bigint (which it does on paper since there is no size limit on numbers).

leighmcculloch: In Stellar JSON APIs we usually use strings for anything over 32bit. Although we haven't done that in the Rust XDR JSON (https://github.com/stellar/xdrgen/issues/181).

The problem is JS not providing anyway to parse numbers into BigInts 😢.

src/contract_spec.ts
```diff
@@ -687,11 +765,323 @@ function findCase(name: string) {
...
+  },
+  I256: {
+    type: 'string',
+    pattern: '^-?[0-9]+$',
+    minLength: 1,
+    maxLength: 79 // Includes additional digit for the potential '-'
+  },
+  Address: {
+    type: 'string',
+    description: 'Address can be a public key or contract id'
```
Shaptic, line 817: does this deserve a regex, as well?

```suggestion
    description: 'Address can be a public key or contract id',
    pattern: '^(G|C)[A-Z0-9]{55}$'
```

Shaptic:
@willemneal is this RFR? I haven't given it a proper look because it's in Draft. Also, can you target the `stable` branch? We should release this alongside the other, final, stable Protocol 20 changes.

willemneal:
I want it to be, but should I move it to js-stellar-sdk? If so which branch should I target there? Is it compatible with testnet?

Shaptic:
@willemneal sorry for the delay. Yeah SDK might be better just so that we don't need to port it over later. There's a `stable` branch to target, see here: https://github.com/stellar/js-stellar-sdk/pull/886

PR #172, opened 2023-11-27
Handle events without contract IDs correctly

Shaptic:
This mirrors https://github.com/stellar/js-stellar-sdk/pull/883.

It also adds a migration guide with justifications and details alongside a deprecation message that people will see when installing `soroban-client`.


sreuland (approved):
lgtm!

